PR #47, opened 2019-09-03
12 pk ratio plot

pchelle:
This is the very first pull request for the PK ratio example.
The fixes of some issues may only be temporary if new class variables are considered.


R/getGrouping.R
```diff
@@ -0,0 +1,24 @@
...
+#' @return data with groups as factor
+#' @export
+#'
+#' @examples
+getGrouping <- function(data, grouping) {
+  # Check that grouping variable are in the data
+  stopifnot(grouping %in% names(data))
+  if (length(grouping)>1){stop("Grouping with more than one variable is not currently handled")}
+  
+  if (is.null(grouping)){
```
msevestre: Code can be simplified a bit. Good to try to always exit early if you can.
```
if(is.null(grouping)){
 return(factor(1))
}

return factor(data...))
```

Limiting the amount of else is crucial to simplify code

R/setFontProperties.R
```diff
@@ -0,0 +1,32 @@
...
+#' @description 
+#' setFontProperties set Font Properties on a ggplot object
+#' @export
+#' @examples
+#' 
+#' p <- ggplot()+labs(title='Title', )
+#' NewFont <- FontClass$new(color="blue", size=20)
+#' p <- setFontProperties(plotHandle=p, TitleFont=NewFont)
+setFontProperties <- function(plotHandle, 
+                              TitleFont=NULL,
```
msevestre: all those objects should be lowercase (titleFont etc..)

msevestre (commented):
@pchelle First review done. I am going to update the code first to be consistent with a package. Then we can discuss about the implementation. Stay tune

R/font-class.R
```diff
@@ -0,0 +1,79 @@
+#' @title Font Class
+#' @docType class
+#' @description
+#' Font class defining properties of a text (size, color, font)
+#' @export
+FontClass <- R6::R6Class(
+  "Font",
+  public = list(
+    "size" = 12,
```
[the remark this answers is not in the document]
pchelle: It seems the elements should be defined in public as well. Initialize seems to be the optional part.
I just tested the initialize without "size" defined in the public, but it did not work.

R/font-class.R
```diff
@@ -0,0 +1,79 @@
...
+                          fontfamily = "",
+                          fontface = "plain") {
+      self$size <- size
+      self$color <- color
+      self$fontfamily <- fontfamily
+      self$fontface <- fontface
+    },
+
+    setFont = function() {
+      element_text(
```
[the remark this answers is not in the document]
pchelle: element_text is a ggplot function that create an element text with color, font, size options for titles, xlabel, ylabel...

R/font-class.R
```diff
@@ -0,0 +1,79 @@
...
+    initialize = function(text = "",
+                          font = NULL) {
+      self$text <- text
+      font <- font %||% FontClass$new()
+
+      stopifnot("Font" %in% class(font))
+      self$font <- font
+    },
+
+    setFontProperties = function(color=self$font$color,
```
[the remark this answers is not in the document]
pchelle: I think set is a predefined function in every R6 class, so I could not use it here

R/font-class.R
```diff
@@ -0,0 +1,79 @@
...
+#' @title asLabel
+#' @param text text to set as label class
+#' @return Label
+#' @description
+#' set a character type into a Label class
+#' @export
+#' @examples
+#' title <- "Title of Plot"
+#' title <- asLabel(title)
+asLabel <- function(text) {
```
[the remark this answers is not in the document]
pchelle: I can create a different script with all these small Label, Fonts related function if it is better ?

R/getGrouping.R
pchelle: Indeed, I set that at the very beginning, and was waiting for Abdullah's code to update.

NAMESPACE
```diff
@@ -1,3 +1,16 @@
 # Generated by roxygen2: do not edit by hand
 
+export(Font)
```
msevestre: @pchelle Time to use the package development
document is created using devtools::document() (CTRL+ALT+D)
package is loaded using CTR+ALT+L

Before submitting code: Run package styling
Run check: CTRL+ALT+E. It will show all erros with the package

R/font.R
```diff
@@ -0,0 +1,80 @@
+#' @title Font
+#' @docType class
+#' @description
+#' Font class defining properties of a text (size, color, font)
+#' @include utils.R
+#' @export
+Font <- R6::R6Class(
```
msevestre: Do not name object with their type (same with file type).

R/pkratio-datamapping.R
```diff
@@ -5,11 +5,26 @@
...
+      self$colorGrouping <- colorGrouping
+      self$sizeGrouping <- sizeGrouping
+      self$shapeGrouping <- shapeGrouping
       super$initialize(x, y)
     }
   )
 )
+
+
+if (FALSE) {
```
msevestre: @pchelle Not sure what this is about here.

R/setLegend.R
```diff
@@ -0,0 +1,135 @@
...
+#' SetLegend set a Legend Position and labels in a plot
+#' @export
+setLegend <- function(PlotHandle, plotConfiguration, dataMapping) {
+
+  # Check Inputs
+  stopifnot(class(PlotHandle) %in% c("gg", "ggplot"))
+
+  # Set Legend Position according to plotConfiguration
+  # plotConfiguration will be added when implemented
+  PlotHandle <- setLegendPosition(PlotHandle)
```
msevestre: @pchelle This should be plotHandle

R/setLegend.R
```diff
@@ -0,0 +1,135 @@
...
+#' @param Location Inside/Oustide/None
+#' @param X.Location NA/Left/Center/Right
+#' @param Y.Location NA/Top/Center/Bottom
+#' @param Subplot.Index (optional) OPTION NOT IMPLEMENTED YET
+#' @return Current plot with Legend Layer
+#' @description
+#' SetLegendPosition set a Legend Position in a plot (case insensitive)
+#' Default legend setting is within plot on the top right corner
+#' @export
+setLegendPosition <- function(PlotHandle,
```
msevestre: @pchelle this should be PlotHandle

R/setWatermark.R
```diff
@@ -0,0 +1,40 @@
...
+
+#' @title createWatermarkGrob
+#' @param label Label class of Watermark
+#' @param angle orientation of the watermark
+#' @return watermark Background as a grob for ggplot objects
+#' @description
+#' createWatermarkGrob use a label to define a Background Watermark as a grob
+#' @export
+createWatermarkGrob <- function(label, angle = 30) {
+  watermark <- ggplot2::ggplot() + ggplot2::theme_void() +
```
msevestre: WIthin a package, you should use ggplot2:: to call ggplot function

R/setWatermark.R
```diff
@@ -0,0 +1,40 @@
+#' @title setWatermark
+#' @param plotHandle ggplot object
+#' @param label to be written in watermark (class Label and character allowed)
+#' @param angle (optional argument) angle of the watermark
+#' @return plotHandle ggplot object with watermark background as a grob
+#' @description
+#' setWatermark set watermark as defined by label
+#' @import  ggplot2
```
msevestre: This is required at least once somewhere to ensure the package is loaded. Might not be required if we declare it as a dependencies. Works like that for now

tests/dev/PKRatioTest.R
```diff
@@ -0,0 +1,78 @@
+# The next line simply remove the variables from the environment
+# Similar to clear from Matlab
+rm(list = ls())
+
+library(ggplot2)
+library(tlf)
```
msevestre: @pchelle This is the only thing required now. No need to source file ever

msevestre (commented):
@pchelle I have added some comments. Remember to always run styler so that we don't have to discuss about formatting

R/font.R
```diff
@@ -0,0 +1,80 @@
+#' @title Font
+#' @docType class
+#' @description
+#' Font class defining properties of a text (size, color, font)
+#' @include utils.R
+#' @export
+Font <- R6::R6Class(
+  "Font",
+  public = list(
+    size = NULL,
```
msevestre: @pchelle This is how I believe you can create your properties and only define the default value in the constructor
Or did you have issues with this approach?

R/font.R
```diff
@@ -0,0 +1,80 @@
...
+    fontfamily = NULL,
+    fontface = NULL,
+
+    initialize = function(size = 12,
+                              color = "black",
+                              fontfamily = "",
+                              fontface = "plain") {
+      self$size <- size
+      self$color <- color
+      self$fontfamily <- fontfamily
```
msevestre: shoud fontFamily and fontFace

Yuri05:
@msevestre Still any TODOs here or can we merge?

pchelle:
I made some modifications based on Michael reviews and I am integrating Abdullah's code into dataMapping and plotConfiguaration.
I wanted to add 1 or 2 examples for the Test folder as well.
I will commit them as soon as possible today, so you can review and merge the fixed code.

msevestre (commented):
Looking good with a few revision added as new issues. Well done

PR #64, opened 2019-10-01
56 default color

pchelle:

PR #66, opened 2019-10-03
Add timeprofile

msevestre:


R/get-pkratio-measure.R
```diff
@@ -25,7 +25,7 @@ getPKRatioMeasure <- function(data, dataMapping = NULL, ratioLimits = c(1.5, 2))
 
   PointsTotal <- length(PKratios)
 
-  PKRatioMeasure <- data.frame("Number" = PointsTotal, "Ratio" = NA, row.names = "PointsTotal")
```
msevestre: Some of those changes are also in #64 that I just reviewed. This is going to be conflicting badly

PR #72, opened 2019-10-10
23 time profile plots

pchelle:


msevestre:
@pchelle The buils fails
https://ci.appveyor.com/project/open-systems-pharmacology-ci/tlf-library/builds/28021305

Can you check it out?

R/datamapping-timeprofile.R
```diff
@@ -5,9 +5,47 @@
...
+      x <- data[, self$x]
+      y <- ifnotnull(self$y, data[, self$y], NULL)
+      yMin <- ifnotnull(self$yMin, data[, self$yMin], NULL)
+      yMax <- ifnotnull(self$yMax, data[, self$yMax], NULL)
+
+      self$data <- as.data.frame(cbind(x, y, yMin, yMax))
+
+      # For each mapped grouping, get the corresponding data frame
+      # Defined by grouping data frame if available or by grouping Name
+      if (!is.null(self$groupings)) {
```
msevestre: What about we initialize groupings to an empty list instead of null? That way, you don't have to always check if it's null
It would make the code probably less defensive

R/datamapping-xygroup.R
```diff
@@ -25,8 +25,7 @@ XYGDataMapping <- R6::R6Class(
 
       if (!is.null(groupings)) {
```
msevestre: here again

R/datamapping-xygroup.R
```diff
@@ -49,7 +48,7 @@ XYGDataMapping <- R6::R6Class(
       # Defined by grouping data frame if available or by grouping Name
       if (!is.null(self$groupings)) {
```
msevestre: and here :)

R/metadata-helper.R
```diff
@@ -0,0 +1,33 @@
...
+#' @title metaDataHelper
+#' @param data data.frame
+#' @return metaData with default OSP Suite units
+#' @description
+#' Create a metaData object with information about units and dimensions of data
+#' @export
+#' @examples
+#' df <- pkRatioData
+#' pkRatioMetaData <- metaDataHelper(data = df)
+metaDataHelper <- function(data) {
```
msevestre: What is this class? Not sure i understand

R/plotconfiguration-timeprofile.R
```diff
@@ -1,42 +1,110 @@
...
+        watermark = watermark,
+        data = data,
+        metaData = metaData,
+        dataMapping = dataMapping
       )
+
+      self$LLOQLinesProperties <- LLOQLinesProperties
+
+      if (!is.null(dataMapping$groupings)) {
+        self$colorGrouping <- dataMapping$groupings[["color"]]$groupingName[[1]]
```
msevestre: It feels that the groupings should be an object with properties color, linetype etc.. Let's discuss

R/plotconfiguration-timeprofile.R
```diff
@@ -1,42 +1,110 @@
...
-      plotHandle <- plotHandle +
-        ifnotnull(self$LLOQLineProperties, ggplot2::geom_hline(
-          yintercept = self$LLOQLineProperties$value,
-          linetype = self$LLOQLineProperties$linetype,
-          color = self$RatioLineProperties$color,
-          size = self$RatioLineProperties$size
-        ))
-      return(plotHandle)
+    addLLOQLines = function(LLOQLines, plotObject) {
+      if (!is.null(LLOQLines)) {
```
msevestre: exit earlyu
if(is.null...)
return

R/plotconfiguration-timeprofile.R
```diff
@@ -1,42 +1,110 @@
...
+      return(plotObject)
+    },
+
+    addTimeProfiles = function(plotObject, data, metaData, dataMapping) {
+      mapData <- dataMapping$getMapData(data, metaData)
+
+      # If no grouping is defined, the aesthetic variables still need to be included
+      # with a dummy name to be removed, otherwise it is impossible to modify them later on.
+      if (is.null(self$linetypeGrouping)) {
+        linetype <- "linetype"
```
msevestre: should it be set to `self$linetypeGrouping` if not null?

R/metadata-helper.R
pchelle: It is a helper function to create a metaData from a dataset.
I don't know if this can be useful moving forward. The idea was to be able to generate easily the metaData of your dataset if you don't have it.
The @examples is  what caused the crash of the build, I am committing the fix now.

R/plotconfiguration-timeprofile.R
pchelle: I have reshaped the grouping quite a few times.
Abdullah's presentation showed there is far more potential for this class compared to what I was expecting at the beginning. It indeed deserve a discussion to figure what we want from this class and consequently how we want it to behave for the API.

PR #75, opened 2019-10-11
Fixes #74 add validaiton helper and Fixes #73

### LINKED ISSUE #73: Localize message into one file so that we do not have to search everywhere for typo

### LINKED ISSUE #74: Add validation helper to enhance validation

msevestre:


R/error-checks.R
```diff
@@ -0,0 +1,87 @@
+isSameLength <- function(...) {
+  args <- list(...)
+  nrOfLengths <- length(unique(lengths(args)))
+
+  return(nrOfLengths == 1)
+}
+
+#' Check if the provided object is of certain type
```
msevestre: Validation function taken right out of OSPSuite.R

R/font.R
```diff
@@ -32,41 +32,3 @@ Font <- R6::R6Class(
...
   )
 )
-
-#' @title Label
-#' @docType class
-#' @description
-#' Label class associate font to a text
-#' @include utils.R
-#' @export
-Label <- R6::R6Class(
```
msevestre: Moved to its own file

R/label.R
```diff
@@ -0,0 +1,34 @@
...
+#' @export
+Label <- R6::R6Class(
+  "Label",
+  public = list(
+    text = NULL,
+    font = NULL,
+
+    initialize = function(text = "", font = NULL) {
+      self$text <- text
+      validateIsOfType(font, Font, nullAllowed = TRUE)
```
msevestre: This is how it's being used. You can pass the type as a string or the type directly
By default, nulI is not allowed but you can specify true when the parameter is optional

R/legend-utils.R
```diff
@@ -0,0 +1,135 @@
...
+#' @param Location Inside/Oustide/None
+#' @param X.Location NA/Left/Center/Right
+#' @param Y.Location NA/Top/Center/Bottom
+#' @param Subplot.Index (optional) OPTION NOT IMPLEMENTED YET
+#' @return Current plot with Legend Layer
+#' @description
+#' SetLegendPosition set a Legend Position in a plot (case insensitive)
+#' Default legend setting is within plot on the top right corner
+#' @export
+setLegendPosition <- function(plotHandle,
```
msevestre: @pchelle We have two methods `setLegentPosition` 
One in `legend-util`, and one in `utilities-legend`. Time to clean that up please. Going to create an issue for this

msevestre: https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/76

R/messages.R
```diff
@@ -0,0 +1,31 @@
+messages <- list(
```
msevestre: @abdullahhamadeh @pchelle 
Please define all string messages in this filee

R/messages.R
```diff
@@ -0,0 +1,31 @@
...
+    callingFunctions <- sys.calls()
+    callingFunction <- sys.call(-length(callingFunctions) + 1)[[1]]
+
+    paste0(
+      callingFunction, ": Arguments '", objectNames,
+      "' must have the same length, but they don't!", optionalMessage
+    )
+  },
+
+  errorEnumNotAllNames = "The enumValues has some but not all names assigned. They must be all assigned or none assigned",
```
msevestre: Example of usage: Advantage. We do not repeat ourselves and all messages are localized so they are easy to change

R/plot-pkratio.R
```diff
@@ -27,8 +27,8 @@ plotPKRatio <- function(data,
     dataMapping = dataMapping
   )
 
-  stopifnot("PKRatioDataMapping" %in% class(dataMapping))
-  stopifnot("PKRatioPlotConfiguration" %in% class(plotConfiguration))
+  validateIsOfType(dataMapping, PKRatioDataMapping)
```
msevestre: This is how it's being used

R/plotconfiguration-axis.R
```diff
@@ -1,3 +1,13 @@
+#' How should comparison of entities be performed
+#'
+#' @include enum.R
+#' @export
+Scaling <- enum(c(
```
msevestre: I have defined the enum Scaling. This should now be used in the code. I did not know how to map it (e.g. I have no idea what identity means). We can discuss about this

abdullahhamadeh:
Yes please re meeting on grouping class.

msevestre:
> Out of curiosity, do we plan to have a unique helper merging the OPSSuiteR messages to these ones ?


Probably not. Some messages will be the same, but some will be context specific.
Plus it seems like a lot of work at the moment to manage yet another package just to share one file. We might get more creative with some git hook eventually. For now, copy :)

msevestre:
>(by the way there was an upper case Scaling in your example, does it mean that the enum variables will start with upper case ?)

the enum DEFINITION is like a class and starts with UpperCase
e.g. Scaling.log,  Scaling.lin etc....

The enum variable, should be lower case 
e.g scaling <- Scaling.log

does it make sense?

pchelle:
It makes perfect sense. Thanks.
Then, I will do the same for legend positions.

PR #77, opened 2019-10-16
76 set legend

pchelle:
I have started integrating the new concepts for groupings.
I have still to work on the integration into the configuration, but the mapping seems already much cleaner


R/datamapping-groupings.R
```diff
@@ -0,0 +1,36 @@
...
+    color = NULL,
+    fill = NULL,
+    linetype = NULL,
+    shape = NULL,
+
+    initialize = function(color = NULL,
+                              fill = NULL,
+                              linetype = NULL,
+                              shape = NULL) {
+      if (!"GroupMapping" %in% class(color)) {
```
msevestre: you can use isOfType here (new helper method)

R/datamapping-groupmapping.R
```diff
@@ -0,0 +1,35 @@
...
+    },
+
+    getCaptions = function(data, metaData = NULL) {
+      if (is.data.frame(self$group)) {
+        captions <- getCustomCaptions(data, self$group)
+      }
+      if (is.character(self$group)) {
+        captions <- getDefaultCaptions(data, metaData, variableList = self$group)
+      }
+      return(captions)
```
msevestre: What happens here if group is neither frame nor string?  This is not complete I believe ?

R/plotconfiguration-axis.R
```diff
@@ -2,9 +2,18 @@
...
+#'  "log", "log10", "log1p", "log2", "logit", "modulus", "probability",
+#'  "probit", "pseudo_log", "reciprocal", "reverse", "sqrt" and "time"
 Scaling <- enum(c(
+  "lin",
+  "log10",
   "log",
-  "lin"
+  "asn", "atanh", "boxcox", "date", "exp", "hms", "identity",
+  "log1p", "log2", "logit", "modulus", "probability",
+  "probit", "pseudo_log", "reciprocal", "reverse", "sqrt", "time"
```
msevestre: !! WOW.

msevestre (commented):
@pchelle A few minor comments. Merging

R/datamapping-groupmapping.R
pchelle: I added a validateIsOfType to allow only the 2 previous options

R/plotconfiguration-axis.R
pchelle: These are all the options available in ggplot, we can select only a few of them if we want

PR #80, opened 2019-10-25
Link vignette

pchelle:
Fix vignette metadata
Fix DESCRIPTION for vignette to work and accounts for other packages (using usethis::use_package() adding fields to suggest)
(I believe it was why the vignettes were not built and linked previously)

When installing package using devtools::install(build_vignettes=TRUE)
It seems to work both by calling vignette("pk-ratio-vignette") or using browseVignettes(package="tlf") 
Let me know if you encounter vignette issues.

Added a draft for release notes


DESCRIPTION
```diff
@@ -23,7 +23,9 @@ LazyData: true
 Suggests: 
     knitr,
     rmarkdown,
-    testthat (>= 2.1.0)
+    testthat (>= 2.1.0),
+    pander,
+    devtools
```
msevestre: I don;t think this is correct. Devtools should not be a dependency as it is PURELY to develop a package. I'd say it's the same for pander

msevestre (changes requested):
I am not sure I understand what's going on here
Why do you need to call devtools in your vignettes?

DESCRIPTION
pchelle: You are right about devtools, it does not need to be in the vignette because it is not used inside the vignette, I just removed it. However, pander is used in the vignette, so it needs to be in the suggested section of namespace.
The documentation: http://r-pkgs.had.co.nz/check.html was quite helpful about the process:
"Checking for unstated dependencies in vignettes. As with tests, every package that you use in a vignette must be listed in the DESCRIPTION. If a package is used only for a vignette, and not elsewhere, make sure it’s listed in Suggests"

PR #83, opened 2019-10-31
Histo branch

abdullahhamadeh:


R/datamapping-xy.R
```diff
@@ -11,17 +11,13 @@ XYDataMapping <- R6::R6Class(
     y = NULL,
     data = NULL,
 
-    initialize = function(x, y) {
+    initialize = function(x, y = NULL) {
```
msevestre: How can Y be null for a datamapping xy? ?

R/plot-histogram.R
```diff
@@ -0,0 +1,66 @@
+set.seed(1234)
```
msevestre: Code that does not belong in the package should not be under R otherwise it will be exported with the package

R/plotconfiguration-histogram.R
```diff
@@ -0,0 +1,143 @@
...
+  public = list(
+    mapData = NULL,
+    bins = NULL,
+    binWidth = NULL,
+    verticalLineFunctions = NULL,
+    verticalLineFunctionNames = NULL,
+
+    #verticalLineProperties = NULL,
+
+    initialize = function(#verticalLineProperties = tlfEnv$currentTheme$verticalLineProperties,
```
msevestre: REmove comment here

R/plotconfiguration-histogram.R
```diff
@@ -0,0 +1,143 @@
...
+  inherit = PlotConfiguration,
+
+  public = list(
+    mapData = NULL,
+    bins = NULL,
+    binWidth = NULL,
+    verticalLineFunctions = NULL,
+    verticalLineFunctionNames = NULL,
+
+    #verticalLineProperties = NULL,
```
msevestre: remove?

R/plotconfiguration-histogram.R
```diff
@@ -0,0 +1,143 @@
...
+      self$bins = bins
+
+      self$verticalLineFunctions = verticalLineFunctions
+      #print( self$verticalLineFunctions )
+      self$verticalLineFunctionNames = verticalLineFunctionNames
+
+    },
+
+
+    addHistograms = function(plotObject,
```
msevestre: I don't like the order of thigs here. Optional parameters should come after mandatory parameters no?

R/datamapping-xy.R
abdullahhamadeh: The histogram takes only an x.  I had the choice of either making a datamapping-x class plus a datamapping-x-group class or reuse datamapping-xy-grouping but give y a null default.   I think the latter might be less cluttered?

R/plot-histogram.R
abdullahhamadeh: Yes this has been replaced

R/datamapping-xy.R
pchelle: Maybe a not relevant remark, for the datamapping-x-group class:
Why not inheriting directly from datamapping-xy-grouping, and then the initialize function won't take y as input ? So, you can call super$initialize with directly y=NULL.
(or x, I was thinking that the plotted variable would be called why in the histogram)

msevestre: I think we should reuse. As long as derived class enfore a y to be present, we are good

PR #86, opened 2019-11-05
Try to add vignettes

msevestre:


msevestre:
@pchelle Even with pandoc installed, it still does not look great. 
I am going to merge this for now and create another issue to investigate
you know where to look now :)

PR #94, opened 2019-11-13
21 box whisker plot

pchelle:
WIP
Draft of boxplot template
Background configuration
Includes cherry pick of population data (to be included in tests and vignettes)


R/datamapping-boxwhisker.R
```diff
@@ -0,0 +1,67 @@
...
+  public = list(
+    quantiles = NULL,
+    outliers = NULL,
+    yMin = NULL,
+    lower = NULL,
+    middle = NULL,
+    upper = NULL,
+    yMax = NULL,
+
+    initialize = function(...,
```
msevestre: @pchelle usually, ... are the last parameters, and not the first. Can we adhere to the convention?

R/datamapping-boxwhisker.R
```diff
@@ -0,0 +1,67 @@
...
+                              middle = NULL,
+                              upper = NULL,
+                              yMax = NULL,
+                              quantiles = c(0.05, 0.25, 0.5, 0.75, 0.95),
+                              outliers = NULL) {
+      super$initialize(...)
+      self$quantiles <- sort(quantiles)
+      self$outliers <- outliers
+
+      self$yMin <- yMin %||% self$quantiles[1]
```
msevestre: Let's create a method that verifies that the array Quantiles has exactly 5 elements
Also what happens if lower and middle is defined but not upper and yMax. Should we enforce either one method or the other?

R/datamapping-boxwhisker.R
```diff
@@ -0,0 +1,67 @@
...
+      self$yMax <- yMax %||% self$quantiles[5]
+
+      self$quantiles <- c(self$yMin, self$lower, self$middle, self$upper, self$yMax)
+    },
+
+    getQuantiles = function(data) {
+      # Redfine group and y while removing NA values
+      group <- ifnotnull(
+        self$x,
+        data[!is.na.data.frame(data[, self$y]), self$x],
```
msevestre: is this correct here? (I don't understand it really but 
for group you do
data[!is.na.data.frame(data[, self$y]), self$x],
and for y you do
data[!is.na.data.frame(data[, self$y]), self$y]

And it's not symetrical. Just want to make sure

msevestre: and it's is correct, this does not need to be evailuated twice

R/datamapping-xygroup.R
```diff
@@ -12,12 +12,12 @@ XYGDataMapping <- R6::R6Class(
...
     },
-    
+
     checkMapData = function(data, metaData = NULL) {
       validateMapping(self$x, data)
-      validateMapping(self$y, data)
+      validateMapping(self$y, data, nullAllowed = TRUE)
 
-      self$data <- data[, c(self$x, self$y)]
+      self$data <- data[, c(self$x, self$y), drop = FALSE]
```
msevestre: This is typically the kind of changes that require an inline comment. This is absolutely not clear why you use drop in this case and not in any other case. We should not comment the obvious, but we should absolutely comment this types of implementation

R/get-boxwhisker-measure.R
```diff
@@ -0,0 +1,67 @@
...
+#' @param ratioLimits vector of numeric
+#' containing the fold limits of PK ratios
+#' As default limits are 1.5 and 2 folds
+#' @description
+#' getPKRatioMeasure(data, dataMapping, ratioLimits) get the number of PK ratios
+#' that are within specific limits
+#' @return a data.frame reporting the PK Ratio Qualification Measure
+#' @export
+#'
+getBoxWhiskerMeasure <- function(data, dataMapping = NULL, y = NULL, group = NULL, quantiles = c(0.05, 0.25, 0.5, 0.75, 0.95)) {
```
msevestre: Please align parameters one per line for consitency

R/get-boxwhisker-measure.R
```diff
@@ -0,0 +1,67 @@
...
+
+  # If no data mapping is input, use default
+  variableNames <- names(data)
+  group <- group
```
msevestre: ??

R/plotconfiguration-boxwhisker.R
```diff
@@ -0,0 +1,88 @@
...
+
+      plotObject <- plotObject + ggplot2::geom_boxplot(
+        data = mapQuantiles,
+        mapping = aes_string(
+          x = mapLabels$x,
+          ymin = mapLabels$yMin,
+          lower = mapLabels$lower,
+          middle = mapLabels$middle,
+          upper = mapLabels$upper,
+          ymax = mapLabels$yMax,
```
msevestre: is there a reason why we are using ` yMax` and not `ymax` (same for ymin).

R/plotconfiguration-histogram.R
```diff
@@ -10,41 +10,24 @@ HistogramPlotConfiguration <- R6::R6Class(
...
-                              title = "Histogram",
+    initialize = function(title = "Histogram",
                               subtitle = paste("Date:", format(Sys.Date(), "%y-%m-%d")),
-                              xlabel = NULL,
-                              ylabel = NULL,
-                              watermark = tlfEnv$currentTheme$watermarkText,
-                              data = NULL,
-                              metaData = NULL,
-                              dataMapping = NULL,
+                              ...,
```
msevestre: last parameter of the initialize function. In the middle is definitely weird

msevestre (changes requested):

R/datamapping-boxwhisker.R
pchelle: I created a new check/error: isOfLength and validateIsOfLength
I am not sure if we should or not: since we have a default quantiles, the user can directly overwrite a part of the boxplot without having to fill the 5 elements each time (unless the user modify quantiles directly).
Thoughts ?

R/datamapping-boxwhisker.R
pchelle: Indeed, I will include the check self$x does not have NA as well.

R/datamapping-boxwhisker.R
msevestre: Yeah I understand
How does that look in intellisense?
for this type of use case,s I would destructure x, and y and add them explicitely before if this is not the case already

R/datamapping-boxwhisker.R
msevestre: in theory it's ok if the quantiles array is null if all other values are undefined. 
This is a weird use case indeed. I am not sure what is best.

R/plotconfiguration-boxwhisker.R
msevestre: yeah let's stick to ggplot nomenclature since we are using everything else from them

msevestre:
@pchelle Build is failing
Let me know when you think this is ready for merging

pchelle:
> 
> 
> @pchelle Build is failing
> Let me know when you think this is ready for merging

I think it is ready :)

PR #120, opened 2020-05-20
104 set plot properties

pchelle:
I apologize for this PR which presents a lot of changes and is still not as good as I want it to. 
But in order to solve many of the reporting engine bugs, the tlf package needs to be updated as frequently as we do with the ospsuite package.

So, if you are okay with this and if the checks are good, we can merge this version and update it with smaller and issue related PRs moving forward.


msevestre:
@pchelle the tlf build is broken now. Can you fix this?

msevestre:
see https://ci.appveyor.com/project/open-systems-pharmacology-ci/TLF-Library/branch/develop

PR #145, opened 2021-06-10
Fixes #144 "blank" shape is ggplot shape 32

### LINKED ISSUE #144: ggplot "blank" shape
`tlf` "blank" shape needs to be `ggplot2` shape 32 instead of -2 to prevent weird characters in Linux

pchelle:


Yuri05:
looks good now! merging

PR #148, opened 2021-07-01
147 plot obs vs pred

pchelle:


DESCRIPTION
```diff
@@ -41,6 +41,7 @@ Suggests:
     gridExtra
 VignetteBuilder: knitr
 Collate: 
+    'aaa.R'
```
Yuri05, line 44: what is aaa.R?

pchelle: Functions that are used to create common R expressions (to prevent copy/paste of code) and that are not exported are stored in the script `aaa.R`; but since these functions are potentially used by any of the remaining functions of the package, I wanted to ensure they are collated first, otherwise build crashes can happen.

I found that other packages used the naming of the files as a solution since they are collated in alphabetical order (e.g. ggplot2 also uses aaa-.R and ospsuite has a zzz.R). 

Would something like `aaa-utilities.R` be more appropriate ?

Yuri05: yes, would be better

Yuri05 (approved):

PR #163, opened 2021-09-15
revamp testing infrstructure - begin

IndrajeetPatil:
# Problem

This is a draft PR that I first wanted to run by you before covering the rest of the tests this way.

Our current approach to testing visual outputs from packages functions is just testing if the returned object is of `ggplot2` class. This is suboptimal since the package functions can return an empty plot or incorrect plot of this class and we wouldn't be able to catch this in CI. 

# Solution

This PR introduces visual regression testing, which adds graphical expectations that can be checked via `{testthat}`.
Additionally, to test `R6` object instances, I am going to use snapshot tests, which are much stricter (because they check the entire object), but also easier to maintain since updating a snapshot requires just running `expect_snapshot()`.

For more about:
- visual regression testing, see: https://vdiffr.r-lib.org/
- snapshot testing, see: https://testthat.r-lib.org/articles/snapshotting.html

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,11 +31,13 @@ Depends:
     R (>= 3.5)
 Encoding: UTF-8
 RoxygenNote: 7.1.2
+Config/testthat/edition: 3
 Roxygen: list(markdown = TRUE)
 Suggests: 
     knitr,
     rmarkdown,
-    testthat (>= 2.1.0),
+    testthat (>= 3.0.0),
+    vdiffr (>= 1.0.0),
     gridExtra
 VignetteBuilder: knitr
 Collate: 
--- a/tests/testthat/test-tornado.R
+++ b/tests/testthat/test-tornado.R
@@ -51,57 +51,57 @@ test_that("TornadoDataMapping features", {
   expect_equal(dataMappingShape$groupMapping$shape$group, "quantile")
 })
 
-if (getRversion() >= "4.1") {
-  test_that("Regular tornado plots with their options work properly", {
-    skip_on_cran()
-    skip_if_not_installed("vdiffr")
+test_that("Regular tornado plots with their options work properly", {
+  skip_on_cran()
+  skip_if_not_installed("vdiffr")
+  skip_if_not(.Platform$OS.type == "windows")
+  skip_if_not(getRversion() >= "4.1")
 
-    # Direct Tornado Plot
-    set.seed(123)
-    defaultPlot <- plotTornado(x = tornadoMeanData$sensitivity, y = tornadoMeanData$path)
-    vdiffr::expect_doppelganger(
-      title = "defaultPlot",
-      fig = defaultPlot
-    )
+  # Direct Tornado Plot
+  set.seed(123)
+  defaultPlot <- plotTornado(x = tornadoMeanData$sensitivity, y = tornadoMeanData$path)
+  vdiffr::expect_doppelganger(
+    title = "defaultPlot",
+    fig = defaultPlot
+  )
 
-    # Options/Features
-    set.seed(123)
-    unsortedPlot <- plotTornado(x = tornadoMeanData$sensitivity, y = tornadoMeanData$path, sorted = FALSE)
-    vdiffr::expect_doppelganger(
-      title = "unsortedPlot",
-      fig = unsortedPlot
-    )
+  # Options/Features
+  set.seed(123)
+  unsortedPlot <- plotTornado(x = tornadoMeanData$sensitivity, y = tornadoMeanData$path, sorted = FALSE)
+  vdiffr::expect_doppelganger(
+    title = "unsortedPlot",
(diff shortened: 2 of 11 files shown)


IndrajeetPatil:
@msevestre, @abdullahhamadeh, @Yuri05, @pchelle What do you think?

I will proceed with the rest of the changes after I hear your thoughts on the current PR.

cc @PavelBal

msevestre:
@IndrajeetPatil my experience with snapshot testing is not so good. It has been used extensively in the javascript community in the past 2-3 years (using jest). The trend now is to move away from this because comparing such tests breaks very often when only `internal details`  from your objects are changing. 

I do not know if the snapshot concept in R is the same as the snapshot concept in Jest/Javascript. If yes, I would only use this if there are absolutely no other alternatives for testing. Otherwise, I am much more in favor of testing the behavior (e.g public API) of an object

DESCRIPTION
```diff
@@ -31,11 +31,13 @@ Depends:
     R (>= 3.5)
 Encoding: UTF-8
 RoxygenNote: 7.1.2
+Config/testthat/edition: 3
 Roxygen: list(markdown = TRUE)
 Suggests: 
     knitr,
     rmarkdown,
-    testthat (>= 2.1.0),
+    testthat (>= 3.0.0),
```
msevestre, line 39: @Yuri05 We can use those packages even if they are not part of the validR env right? It's just for testing so not a direct dependencies

msevestre (commented):
@IndrajeetPatil I'd love to meet so that you can explain what's going on here. It's too new for me to be able to understand . Probably good for everyone in the team

DESCRIPTION
IndrajeetPatil: Yes, absolutely!

I would be happy to chat about this in person and maybe also do a live demo of these testing frameworks and how they would play out for our packages.

IndrajeetPatil:
> The trend now is to move away from this because comparing such tests breaks very often when only internal details from your objects are changing.

100% true. These tests are quite brittle, but the way they are implemented in R, it is also very easy to modify them.

That said, after our in-person talk, we can decide whether to have them as an alternative or as a companion to the other behaviour-based testing

Yuri05:
I was wondering how robust the tests are and have run them under Linux (CentOS7) with R 3.6.3.
It seems that R 3.6 is not supported?
grafik

IndrajeetPatil:
Ah, yes.

The R graphics engine changed in 4.0 release, and so the visual regression tests can only be run on release and development version of R. I have 4.1.1 on my local computer, so I could run them. 

I think we should (although I am not sure how easy would be) check the package on Appveyor for three different releases of R: oldrel, release, and devel, if we aren’t doing so already. 

Btw, ggplot2 also uses these tests to checks its outputs so vdiffr is the standard for visual regression testing in R.

IndrajeetPatil:
> @IndrajeetPatil I'd love to meet so that you can explain what's going on here. It's too new for me to be able to understand . Probably good for everyone in the team

Yes, absolutely!

I would be happy to chat about this in person and maybe also do a live demo of these testing frameworks and how they would play out for our packages.

Yuri05:
>The R graphics engine changed in the 4.1 release 

Ok, that means we will not be able to run those new tests within the current ValidR-environment (3.6.3-based) and even within the next ValidR (4.0.4-based). If we keep the tests, we will need to disable them on Linux then for now.

IndrajeetPatil:
> If we keep the tests, we will need to disable them on Linux then for now.

Yes, we can easily skip them on a specific OS.

PavelBal:
> Ok, that means we will not be able to run those new tests within the current ValidR-environment (3.6.3-based) and even within the next ValidR (4.0.4-based). If we keep the tests, we will need to disable them on Linux then for now.

Why is it required to run the tests in ValidR?

Yuri05:
>Why is it required to run the tests in ValidR? 

Because it's part of our standard package validation procedure.

tests/testthat/test-tornado.R
```diff
@@ -47,15 +52,44 @@ test_that("TornadoDataMapping features", {
 })
 
 test_that("Regular tornado plots with their options work properly", {
+  skip_on_cran()
+  skip_if_not_installed("vdiffr")
+  skip_if_not(.Platform$OS.type == "windows")
+  skip_if_not(getRversion() >= "4.1")
```
Yuri05, line 58: I think we can remove `skip_if_not(.Platform$OS.type == "windows")`
When I said "skip on Linux" I had the Linux ValidR platform in mind. 
But having 2 other conditions (vdiffr and R>=4.1) is even better and should be enough in this case (and as soon as the ValidR-platform supports it - the tests should run there as well)

IndrajeetPatil: Got it! Thanks for the clarification. I have reverted that change.

PR #188, opened 2021-11-03
187 histograms

pchelle:


R/plot-boxwhisker.R
```diff
@@ -18,7 +18,7 @@
 #' @return a ggplot graphical object
 #'
 #' @references For examples, see:
-#' <https://www.open-systems-pharmacology.org/TLF-Library/articles/box-whisker-vignette.html>
+#' <https://www.open-systems-pharmacology.org/TLF-Library/articles/box-whisker.html>
 #'
```
IndrajeetPatil, line 22: The previous link was correct. If you go to this new link, it's 404. 

Note that the URLs are created based on the names of the Rmd files, and the name of this vignette is indeed `box-whisker-vignette.Rmd`:
https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/develop/vignettes/box-whisker-vignette.Rmd

So please revert this change.

PR #197, opened 2021-11-09
adds github actions

IndrajeetPatil:


Yuri05:
@IndrajeetPatil @msevestre With the new code coverage GiHub action the usage of code coveage can be removed from AppVeyor.yml, right?
https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/aea83b90ac0fd2dcb75c30889852edaaa8f8f6a7/appveyor.yml#L54-L60

IndrajeetPatil:
Yes, I think that should be okay.

The R CMD Checks are failing though because the vignettes can't be knitted properly.

.github/workflows/check-pak.yaml
```diff
@@ -0,0 +1,67 @@
...
+          - {os: macOS-latest,   r: 'release'}
+
+          - {os: windows-latest, r: 'release'}
+          # Use 3.6 to trigger usage of RTools35
+          - {os: windows-latest, r: '3.6'}
+
+          # Use older ubuntu to maximise backward compatibility
+          - {os: ubuntu-18.04,   r: 'devel', http-user-agent: 'release'}
+          - {os: ubuntu-18.04,   r: 'release'}
+          - {os: ubuntu-18.04,   r: 'oldrel-1'}
```
Yuri05: "release" is the latest released r-version (like 4.1.2 now), "devel" is the next development version (e.g. 4.1.3 or 4.2.0)... and what is "oldrel-1"? 4.1.1 or 4.0.x or 3.6.5?

IndrajeetPatil: devel is `4.2.0`

Rest are these versions:

``` r
library(rversions)

r_oldrel()
#>     version                date        nickname
#> 122   4.0.5 2021-03-31 07:05:15 Shake and Throw

r_release()
#>     version                date    nickname
#> 125   4.1.2 2021-11-01 08:05:12 Bird Hippie
```

<sup>Created on 2021-11-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

IndrajeetPatil:
I am gonna skip the R CMD Check GHA for now, and come back to it later. The code coverage, linting, and pkgdown (🤞 ) actions should work.

msevestre:
Stupid question maybe: Why are we updating the way code coverage is performed? What is the benefit?

msevestre:
specificaly we are not using our special secret key for coveralls. 

  COVERALLS_TOKEN:
    secure: QItC20vr6rKHkiDVvsE+kHFL4GTS3bcsPw5KcWe+qipEIj1+ncKngsCoCNelNdtg

IndrajeetPatil:
No added benefit, except now we get everything in one place:

GHA gives us           : code coverage + website + lint detection + R CMD Check
Appveyor provides  : code coverage + R CMD Check

> specificaly we are not using our special secret key for coveralls

We don't need to do this. GHA uses codecov and uses secret GitHub PAT provided for our repo, so we don't need to set anything. 

Are you set on using coveralls, or codecov is also fine? Both of them provide identical information.

IndrajeetPatil:
That said, I am not wedded to the idea. Given the complexities of dotnet wrappers, etc.

We can use:
GHA only for website,
AppVeyor for R CMD Check and code coverage

That's also fine by me.

IndrajeetPatil:
Holy shit! pkgdown GHA works successfully! 

So, from now once, the website will update itself after each commit!! 🥳 

@msevestre Can you please change settings/pages to the following:

image

msevestre:
I was just wondering what was the *benefit* of migrating to this new GHA (I do not know the syntax so the 30 lines of yaml look a bit scary to me) where we had one line before.
At any rate, looks good to me. And codecov is enough. Let's update the appveyor.yml to remove the codecoverage stuff
We should so the same for our other repositories for consistency sake

msevestre:
done
Seems to be only in darkmode now? is it on purpose?

msevestre:
correction 
http://www.open-systems-pharmacology.org/TLF-Library/
is in darkmode
buts
https://www.open-systems-pharmacology.org/TLF-Library/ is outdated

(see HTTP vs https)

IndrajeetPatil:
Strange. This is not a pkgdown website. Something is amiss.

What is the URL given by GitHub pages (`settings/pages`) after `Your site is published at ....`

IndrajeetPatil:
And, to be absolutely sure, we are using `/(root)` directory in `gh-pages` branch, right? Not `/docs/`?

image

msevestre:
@IndrajeetPatil You are now admin..... enjoy:)

IndrajeetPatil:
Thanks!

We just needed to enforce HTTPS protocal. 

The website is all set now and can be seen here: https://www.open-systems-pharmacology.org/TLF-Library/

The look of the website can be changed using the parameters here:

https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/f6e5f9ce1b415158a82fdea00889bd3f6236fb84/_pkgdown.yml#L3-L6

I am obseesed with dark mode, so that's why I have chosen. But feel free to change these later.

msevestre:
I don't mint dark mode but this is really "in your face" dark mode lol
I think for now we should keep all page the same. (e.g. white or whatever we have for other packages) and go from there.

Also is the me or the plots are broken?
image

IndrajeetPatil:
Okay, I have turned off the dark mode now. 

I see the same output if I knit the vignette locally, so I don't think it has anything to do with pkgdown or GHA. 
Those examples are broken in the vignette itself.

Yuri05:
@IndrajeetPatil Cool, well done! 
And in case we have a voting on the dark mode: I'm against it :) Prefer the "normal" mode 100%

IndrajeetPatil:
"normal" 🤣

PR #204, opened 2021-11-26
fixes #203 Add artifacts without version number to enable usage as pe…

### LINKED ISSUE #203: Add artifacts without version number to enable usage as permalink
Create additional artifacts **tlf.zip** and **tlf.tar.gz** (without the version number), so that they can e used in the permalinks to the latest built package version (s. the discussion in https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/pull/592)

Yuri05:
…rmalink

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -56,10 +56,12 @@ build_script:
 test_script:
   - travis-tool.sh run_tests
 
-on_success:
-  - Rscript -e "covr::codecov()"
+after_test:
   - ps: copy \tlf_*.tar.gz \tlf.tar.gz
   - ps: copy \tlf_*.zip \tlf.zip
+
+on_success:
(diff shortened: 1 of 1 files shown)


msevestre:
I don't think you can install such a package then. At least the zip file on Windows. I had tried before and it said the package was corrupted

Yuri05:
>I don't think you can install such a package then. At least the zip file on Windows. I had tried before and it said the package was corrupted

I used it in the latest PR on RE - seems to work so far

msevestre:
Can you install it locally using r studio?
I think we should also have the version with version if not

Yuri05:
>I think we should also have the version with version if not

Yep. Now we have both artifacts: with and without the version (a **copy** of created package is created before publishing the artifacts)
grafik

PR #213, opened 2021-12-20
209 use ospsuite utils

pchelle:


pchelle:
About the use of `::` and the difference between Namespace and Imports, I found this: https://r-pkgs.org/namespace.html#imports interesting

From what I understand, the use of `::` is especially advised when there is a chance that the function or object name is not unique or was updated by another package (e.g. `base::sample` vs `dplyr::sample`).
Since one of the goals of `{ospsuite.utils}` is to prevent that by centralizing the code, I think it is okay not to specify `ospsuite.utils::` in most cases.

Yuri05 (approved):

PR #244, opened 2022-04-21
234 alignment

pchelle:


R/font.R
```diff
@@ -1,54 +1,106 @@
...
+#' @export
+Font <- R6::R6Class(
+  "Font",
+  public = list(
+    size = 12,
+    color = "black",
+    fontFamily = "",
+    fontFace = "plain",
+    angle = 0,
+    align = "center",
```
Yuri05, line 18: let's set the default alignment to "left". seems more convenient

Yuri05 (approved):

Issue #241, opened 2022-04-13
Cannot install ospsuite.utils


Yuri05:
https://github.com/Open-Systems-Pharmacology/TLF-Library/runs/6004849722?check_suite_focus=true#step:4:48
```
Error: Error: <callr_remote_error: Cannot install packages:
  * local::.: Can't install dependency ospsuite.utils
  * ospsuite.utils: Can't find package called ospsuite.utils.>
   in process 8537 
  -->
  <simpleError: Cannot install packages:
  * local::.: Can't install dependency ospsuite.utils
  * ospsuite.utils: Can't find package called ospsuite.utils.>
```

IndrajeetPatil:
This is because we archived _ospsuite.utils_ from CRAN (cf. https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/issues/106#issuecomment-1089200141), and so now it's available only on GitHub.

Trying a few things here: https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/242

IndrajeetPatil:
We have removed GHA from our workflows now.

PR #246, opened 2022-04-23
move plot grid function and class from ospsuite

IndrajeetPatil:
cf. https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/888


IndrajeetPatil:
Not sure, why there are these issues with installing pandoc now.

```r
Failures
 - pandoc (exited 1) - pandoc not installed. An error occurred during installation:
 The remote server returned an error: (429) Too Many Requests. Too Many Requests
Command exited with code 1
7z a failure.zip *.Rcheck\*
7-Zip 21.07 (x64) : Copyright (c) 1999-2021 Igor Pavlov : 2021-12-26

```

Yuri05:
"Too Many Requests" probably server too busy. 
I restarted the build manually and it seemed to work (at least with pandoc) - so was just a one time glitch

R/plot-grid.R
```diff
@@ -0,0 +1,170 @@
...
+#'   of rows, columns, labels, etc.).
+#'
+#' @description
+#'
+#' Create a plot grid using the `patchwork::wrap_plots()` function. The required
+#' arguments are supplied through the `PlotGridConfiguration` object.
+#'
+#' @examples
+#'
+#' library(ospsuite)
```
Yuri05, line 14: remove `#' library(ospsuite)`

R/plot-grid.R
```diff
@@ -0,0 +1,170 @@
...
+    )
+}
+
+
+#' @title Class for creating a plot grid
+#'
+#' @description
+#'
+#' R6 class defining the configuration for `{patchwork}` plot grid used to
+#' create a grid of plots from `{ospsuite}`. It holds values for all relevant
```
Yuri05, line 101: remove `from `{ospsuite}`` or replace it with `from `{tlf}``

man/plotGrid.Rd
```diff
@@ -0,0 +1,75 @@
...
+is an \code{R6} class object that defines properties of a plot grid (like number
+of rows, columns, labels, etc.).}
+}
+\description{
+Create a plot grid using the \code{patchwork::wrap_plots()} function. The required
+arguments are supplied through the \code{PlotGridConfiguration} object.
+}
+\examples{
+
+library(ospsuite)
```
Yuri05, line 20: remove` library(ospsuite)`

IndrajeetPatil:
@Yuri05 Removed the references to _ospsuite_, which were left in by mistake.

Yuri05 (approved):

msevestre:
appveyor works but not test coverage.

IndrajeetPatil:
@msevestre Yes, this is same as this issue: https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/241

GHA is no longer working. We either need to fix it (I unsuccessfully tried in #242) or completely remove it.

msevestre:
Can we do code coverage like we do for all other packages? Tha way we don't rely on it anymore. We keep having issue with GHA and that's a real PIA

IndrajeetPatil:
Well, we are having issues now because we removed _ospsuite.utils_ from CRAN, but it was working fine before.

Actually, the test coverage GHA is redundant because we are already using AppVeyor for this. 
In fact, for this very PR, code coverage was successfully computed (https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/246#issuecomment-1107532000). So we can just remove GHA for test-coverage.

Yuri05:
>Well, we are having issues now because we removed ospsuite.utils from CRAN, but it was working fine before.

Any problems to install ospsuite.utils in GHA in the same way as in appveyor.yml?
https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/dd00853dd2a82c7a61daca9e400e3079e5e3fc6d/appveyor.yml#L55

>Actually, the test coverage GHA is redundant because we are already using AppVeyor for this.
In fact, for this very PR, code coverage was successfully computed (https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/246#issuecomment-1107532000). So we can just remove GHA for test-coverage.

I would prefer the other way around: remove code coverage from appveyor.yml but keep the GitHub action.
Because AppVeyor and GHA are executed in parallel. And in some repositories (e.g. Reporting Engine) Code coverage takes ages - so executing it in parallel would reduce the build time.
Or are there any advantages of code coverage in appveyor (@msevestre ) ?

msevestre:
Code coverage doesn't have to be run each time. I think we need to split it into it's own task (or have it work with GA). In general, code coverage gives you some nice metrics but isn't required to decide whether to merge a PR or not. So as for our main apps(pksim, mobi and core), I would suggest to create a specific task for this that would run on a weekly basis, especially if it takes a lot of time. Even with GHA, this is time that we do need to wait for everytime to get valuable CI feedback

Yuri05:
agree. running codecovr nightly/weekly/whatever is surely better than on every commit

PR #266, opened 2022-05-12
Fixes #263 add enum helpers for units and format

### LINKED ISSUE #263: Create enums relevant for `ExportConfiguration`
In the following class:

```r
ExportConfiguration$new(
  name = NULL,
  format = NULL,
  width = NULL,
  height = NULL,
  units = NULL,
  dpi = NULL
)
```

I am assuming the supported file formats are the same ones supported by _ggplot2_?

> one of "eps", "ps", "tex" (pictex), "pdf", "jpeg", "tiff", "png", "bmp", "svg" or "wmf" (windows only)

If yes, I think we can create enum for the same.

Similarly, we can also create an enum for units specifying plot dimensions:

> Plot size in units ("in", "cm", "mm", or "px")

pchelle:


IndrajeetPatil (approved):
Perfect! Thanks.

PR #270, opened 2022-05-13
262 res vs pred

pchelle:


R/plot-res-vs-pred.R
```diff
@@ -78,3 +82,31 @@ plotResVsPred <- function(data,
...
+#' 
+#' plotResVsTime(data = resVsTimeData, dataMapping = ResVsTimeDataMapping$new(x = "x", y = "y"))
+#' 
+#' # Produce Res vs Time plot with linear regression
+#' plotResVsTime(
+#' data = resVsTimeData, 
+#' dataMapping = ResVsTimeDataMapping$new(x = "x", y = "y"),
+#' smoother = "lm"
+#' )
+plotResVsTime <- plotResVsPred
```
Yuri05, line 112: I don't understand this.
Here `plotResidualsVsTime `function is just set equal to `plotResidualsVsPredicted `function?
So calling it would produce again residuals vs predicted plot and not residuals vs time plot???

R/plotconfiguration.R
```diff
@@ -333,6 +333,11 @@ ResVsPredPlotConfiguration <- R6::R6Class(
   inherit = PlotConfiguration
 )
 
+#' @title ResVsTimePlotConfiguration
+#' @description R6 class defining the configuration of a `ggplot` object for Res vs Pred/Time plots
+#' @export
+ResVsTimePlotConfiguration <- ResVsPredPlotConfiguration
```
Yuri05, line 339: same question as for `ResVsTimeDataMapping `

R/plot-res-vs-pred.R
pchelle: Before users had to use `plotResVsPred` and 
- define `x` = predicted in their dataMapping if the plot was actually residuals vs predicted.
- define `x` = time in their dataMapping if the plot was actually residuals vs time.

Having the exact same function for `plotResVsTime` does not change this workflow.
Users still need to map the `x` variable, which drives if the actual plot is residuals vs time or predicted.
For this reason, I did not think there is a need to "copy/paste" the function.
Using an alias seems a better solution to provide a function with a name consistent to what the plot display.

If you think this might be more confusing in the end, I can duplicate the code

Yuri05: no, no code duplications pls :)
ok, so in fact we have just an XY-Scatter-Plot function with different "aliases" (plotResVsPred, plotResVsTime)?
Well, why not.

pchelle: Yes, they are both XY scatter plots + a line at y=0 + optional smoother/regression

msevestre (changes requested):
I don't think this is right for class. Please create another class that derives from this one instead.

R/plotconfiguration.R
msevestre: I don't think this is right for class. Please create another class that derives from this one instead.

R/obs-vs-pred-datamapping.R
```diff
@@ -35,3 +35,9 @@ ResVsPredDataMapping <- R6::R6Class(
   "ResVsPredDataMapping",
   inherit = ObsVsPredDataMapping
 )
+
+#' @title ResVsTimeDataMapping
+#' @description  Class for mapping variables in residuals vs predictions/time plot
+#' @export
+#' @family DataMapping classes
+ResVsTimeDataMapping <- ResVsPredDataMapping
```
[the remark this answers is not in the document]
msevestre: agreed with Juri
Let's create a derived class for this that has no specific implementation

Or better. We find a name that makes sense for both use cases

IndrajeetPatil:
The PR has conflicts now that need to be resolved.

IndrajeetPatil:
@pchelle Can you please resolve the merge conflicts here. 

I'd like to merge this before #305 to reduce likelihood of another merge conflict.

IndrajeetPatil:
@pchelle Build is failing after the conflicts were resolved.

IndrajeetPatil (approved):
Alright, everything looks good now.

Thanks, @pchelle!

PR #271, opened 2022-05-16
Fixes #264 Avoid manual update for the caption of time profile plot

### LINKED ISSUE #264: FR: Avoid manual update for the caption of time profile plot
It's not clear to me as to why, but, the caption for `tlf::plotTimeProfile()` output needs to be manually updated using `setLegendCaption()`. This is quite tricky and error-prone. 

Is there any way that this is internally handled by the plotting function itself? 

It is a bit surprising that the caption is not automatically updated to the chosen mappings for the plot, and needs to be updated manually.

pchelle:


R/plot-timeprofile.R
```diff
@@ -257,6 +257,7 @@ plotTimeProfile <- function(data = NULL,
     getAestheticValues(
       n = totalLength - colorLength,
       selectionKey = plotConfiguration$points$color,
+      position = colorLength,
       aesthetic = "color"
```
msevestre, line 261: Setting a position to a number of colors is not obvious. This needs to be commented

msevestre (changes requested):

IndrajeetPatil:
@pchelle Thanks! 

I can confirm that (in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/956) that this fixes #264.

PR #295, opened 2022-05-24
adopt same gitattributes as other R repos in OSP

IndrajeetPatil:


msevestre:
I read adopt some "gratitude" lol

PR #296, opened 2022-05-25
Add `collate` directive

IndrajeetPatil:
To avoid any headaches while using package objects (enums mostly) in R6 class environments, we need to have a `Collate` directive that makes these objects available before we come to the class definition files.


IndrajeetPatil:
There might be some more whitespace changes due to #295

DESCRIPTION
```diff
@@ -1,100 +1,100 @@
...
+Encoding: UTF-8
+RoxygenNote: 7.2.0
+Roxygen: list(markdown = TRUE)
+Suggests:
+    knitr,
+    rmarkdown,
+    testthat (>= 3.0.3),
+    vdiffr (>= 1.0.0)
+VignetteBuilder: knitr
+Collate:
```
PavelBal, line 47: Are you using roxygen to create the collate field, or do it manually?

IndrajeetPatil: Roxygen

That's why I have added `#' @include` directive which was missing before. That's what created problems.

PavelBal (approved):

PR #293, opened 2022-05-24
Improvements to the `PlotGridConfiguration` class

IndrajeetPatil:
- [x] Closes #274 
- [x] Closes #294

PR #297, opened 2022-05-25
Also add watermark size to the enum

IndrajeetPatil:
Missed this one in #293


IndrajeetPatil:
Merging as the build failure is related to Pandoc, and it's a really minor change.

PR #299, opened 2022-05-27
Use utilities from ospsuite.utils

IndrajeetPatil:
To be considered once https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/pull/115 is merged.


IndrajeetPatil:
Yeah, nothing fails because these functions don't seem to be used anywhere in the package.

PR #305, opened 2022-06-13
Fixes #304 add enum helper for ticklabels

### LINKED ISSUE #304: Create an enum for `TickLabelTransforms`
Should return a character which when passed to internal _tlf_ functions will create a function. 

For example,

```r
tlf::TickLabelTransforms$identity
#> "identity"
```

which is then internally converted to:

```r
function(x) x
```

pchelle:


IndrajeetPatil (approved):
Tried the new functions in _ospsuite_ plot config, and it works as expected. Thanks!

PR #316, opened 2022-06-23
Fixes #312 remove infinite values from plots

### LINKED ISSUE #312: Dealing with 0s on obs vs pred plot
In logarithmic scale, I think these values can be removed (with a warning) instead of displaying them as OOB.

174253122-0102f21b-436a-4d9a-968a-b6bcd1fe62a1

pchelle:


R/plotconfiguration-axis.R
```diff
@@ -287,7 +287,8 @@ XAxisConfiguration <- R6::R6Class(
             trans = self$ggplotScale(),
             breaks = self$prettyTicks(),
             labels = self$prettyTickLabels(),
-            expand = self$ggplotExpansion()
+            expand = self$ggplotExpansion(),
+            oob = .removeInfiniteValues
```
Yuri05, line 291: `oob = .removeInfiniteValues` or
`oob = .removeInfiniteValues() `?

pchelle: `oob` argument expects a `function`. So, `oob = .removeInfiniteValues` is good
> Another way of doing this could be `match.fun(".removeInfiniteValues ")`

The default value is `oob = censor`, passing the function `censor` from the `{scales}` package. 
However , the function does not remove `Inf` values by default.

Interestingly, all the arguments of `scale_y_continous` function handle `function` type values.
This was actually the approach for solving the ticks and tick labels of log scale plots 
> `labels = self$prettyTickLabels()` returns functions such as `getLogTickLabels` for log scale plots, but can also return `getPiTickLabels` if you are working with angles in radians or `getGreekTickLabels` if you want to use greek letters to name indices (these functions were refined in PR #304 and we could also envision a function for setting a calendar time format for tick labels)

Yuri05: interesting :)

PR #328, opened 2022-07-06
Fixes #325 legend title properties from argument title are used

### LINKED ISSUE #325: Legend title not updated with properties specified in `Label` object
As discussed in the meeting.

pchelle:
Issue was caused because properties of R6 (Label object here) are linked if object is not cloned


IndrajeetPatil:
Thanks; can confirm that indeed solves the issue.

PR #331, opened 2022-07-12
Fixes #327 horizontal bars for obs vs pred plots

### LINKED ISSUE #327: Adding horizontal error bars to `plotObsVsPred()`
As specified [here](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/945#issuecomment-1173770611), is it possible to add a horizontal error bar for observed data in `plotObsVsPred()`? 

If this is currently not possible, can we support this?

image

pchelle:
- error is replaced in favor of ymin/ymax or xmin/xmax
- dataMapping initialize method needs to explicitly redefine `x` and `y` arguments because of R partial matching
> if user inputs `x="a"`, partial matching would assign "a" to `xmin` instead of `x` because `xmin` was the only argument explicitly defined

### DIFF
--- a/R/aaa-utilities.R
+++ b/R/aaa-utilities.R
@@ -152,25 +152,43 @@
 #' @description Create an expression that adds errorbars if uncertainty is included in dataMapping
 #' @return An expression to `eval()`
 #' @keywords internal
-.parseAddUncertaintyLayer <- function() {
-  expression({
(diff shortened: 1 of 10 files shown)


IndrajeetPatil:
Tried this in _ospsuite_, and can confirm that this works as expected! Thanks a lot.

IndrajeetPatil (approved):

PR #335, opened 2022-07-18
Refresh README

### LINKED ISSUE #153: Adjust readme
I think we need a proper readme here, with the brief description of the package.

### LINKED ISSUE #334: Fix badges in `README`
The `{pkgdown}` attributes are included in `README`. They should actually be included in `.Rmd`, and not `.md` file.

image

IndrajeetPatil:
Closes #153
Closes #334


IndrajeetPatil:
You only need to review `README.Rmd`, and not `README.md`.

Yuri05:
@IndrajeetPatil Why do we need to generate README.md from README.Rmd?
The advantage of Rmd is R code execution during the generation process - but here we don't have any R code to be executed...

IndrajeetPatil:
How can you solve #334 without using `.Rmd`?

IndrajeetPatil:
I mean now the issue is solved, but it wouldn't have been without the Rmd. 

So I can delete it now, but, in general, I prefer generating md from Rmd.

Yuri05:
> How can you solve #334 without using `.Rmd`?

Is the pkgdown Intro page generated from README.md or from README.Rmd?
If from .Rmd - don't know how to solve it without :)

PavelBal (approved):

IndrajeetPatil:
Yeah, I have removed the .Rmd file now.

README.md
```diff
@@ -1,43 +1,72 @@
-# TLF Library
 
-TLF Library implementation in R
+<!-- README.md is generated from README.Rmd. Please edit that file -->
```
Yuri05, line 2: let's remove the comment as well

IndrajeetPatil: Done.

Yuri05 (approved):

PR #338, opened 2022-07-18
Get rid of warnings and notes in R CMD check

### LINKED ISSUE #162: get rid of warnings and notes in package check
## NOTES

```r
N  checking for hidden files and directories ... 
   Found the following hidden files and directories:
     scripts/.git
     .github
   These were most likely included in error. See section 'Package
   structure' in the 'Writing R Extensions' manual.

N  checking top-level files ...
   Non-standard files/directories found at top level:
     '_pkgdown.yml' 'rakefile.rb' 'scripts'

N  checking R code for possible problems (11.9s)
   mergeLegend: no visible binding for global variable 'color'
   mergeLegend: no visible binding for global variable 'shape'
   mergeLegend: no visible binding for global variable 'size'
   mergeLegend: no visible binding for global variable 'linetype'
   mergeLegend: no visible binding for global variable 'fill'
   plotTimeProfile: no visible binding for global variable 'fillLength'
   plotTimeProfile: no visible binding for global variable
     'linetypeLength'
   setCaptionColor: no visible binding for global variable 'newPlotObject'
   setCaptionFill: no visible binding for global variable 'newPlotObject'
   setCaptionLabels: no visible binding for global variable
     'newPlotObject'
   setCaptionLinetype: no visible binding for global variable
     'newPlotObject'
   setCaptionOrder: no visible binding for global variable 'newPlotObject'
   setCaptionShape: no visible binding for global variable 'newPlotObject'
   setCaptionSize: no visible binding for global variable 'newPlotObject'
   setCaptionVisibility: no visible binding for global variable
     'newPlotObject'
   Undefined global functions or variables:
     color fill fillLength linetype linetypeLength newPlotObject shape
     size
```

## WARNINGS

```r
W  checking dependencies in R code (1.7s)
   '::' or ':::' import not declared from: 'shiny'
```

IndrajeetPatil:
Closes #162


R/global-vars.R
```diff
@@ -0,0 +1,17 @@
...
+    "fillLength",
+    "linetype",
+    "linetypeLength",
+    "newPlotObject",
+    "shape",
+    "size"
+  ),
+  package = "tlf",
+  add = FALSE
+)
```
Yuri05, line 17: I was trying to understand where those global varibales are defined - and failed.
E.g. "fillLength": 
https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/87f42c2c606cca31c7f4f055ca9b8a91868a2ced/R/plot-timeprofile.R#L104-L109
The comment says "fillLength defined in .parseUpdateAestheticProperty" - but there is no such definition in `.parseUpdateAestheticProperty` - and also not elsewhere in the code.
https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/87f42c2c606cca31c7f4f055ca9b8a91868a2ced/R/aaa-utilities.R#L84-L98
I am very confused

IndrajeetPatil: I think that comment is outdated. 

As mentioned in this NOTE:

plotTimeProfile: no visible binding for global variable 'fillLength'

This unquoted symbol or name appears in the code but it is not defined anywhere, and so R CMD Check complaints.

This is a common problem with non-standard evaluation and I am using the standard solution.

IndrajeetPatil: Ref: https://r-pkgs.org/check.html#sec-check-checks

image

Yuri05: Well, maybe it's just my very limited understanding of R, but I just not get how you can use undefined variable? It must be defined somewhere, or not?

Yuri05: ok, can you guide me to where `fillLength `is defined in the code?
Trying to find it with the full search, I can only find 3 usages: 2 in the functions **plot-boxwhisker** and **plot-tornado** (where `fillLength ` it's just a "normal" local variable with explicitely assigned value) and finally the last usage in **plot-timeprofile**, where I don't understand where the value for `fillLength ` comes from) 
grafik

IndrajeetPatil: Sorry, I am not gonna to be of much help here. I am not familiar with this code, so @pchelle can explain this better than I can.

IndrajeetPatil: IINM, he is on vacation this week.

pchelle: Sorry for the delay.
The variable `fillLength` is created during the call of `.parseUpdateAestheticProperty` when input argument is `"fill"`
To prevent potential copy/paste issues of very similar lines of code, most of the functions from `aaa-utilities.R` directly output R code as character strings (which are then transformed into expressions to be evaluated as actual R code using `parse(text)` and `eval(expression)`.

Below shows the expression returned when the function `.parseUpdateAestheticProperty` is called with the arguments "fill" and `ribbons`. 
```r 
> .parseUpdateAestheticProperty("fill", "ribbons")

expression(
fillVariable <- gsub("`", "", mapLabels$fill),
fillLength <- length(unique(mapData[, fillVariable])),
suppressMessages(
plotObject <- plotObject + ggplot2::scale_fill_manual(
values = .getAestheticValues(n = fillLength, selectionKey = plotConfiguration$ribbons$fill, aesthetic = "fill")
)
), 
if (isIncluded(fillVariable, "legendLabels")) {
plotObject <- plotObject + ggplot2::guides(fill = "none")
}
)
```

When the expression is wrapped by `eval()`, this code is evaluated and 
- Creates the variable `fillVariable`, column name in the data linked to the `fill` property
- Creates the variable `fillLength`, number of unique values for the `fill` property (used for selecting the right number of colors)
- Update the `fill` properties of the plot `plotObject`
- Remove the legend if no mapping was actually defined

Note that changing `"fill"` by any other property name such as `"color"` or `"alpha"` would directly create an evaluated R code with all the property names changed to `"color"` or `"alpha"`

pchelle: I think the warning in R-CMD-Check is from the function `plotTimeProfile` which needs to manage between the aesthetic properties of errorbars, scatter points, ribbons and lines while potentially sharing the `color` aesthetic property.

The `plotTimeProfile` code was leveraging the creation of the `fillLength` property created by a previous call of `.parseUpdateAestheticProperty` to merge the `fill` property of both simulated and observed data.

Since there is no actual written code of `fillLength <- length(unique(mapData[, fillVariable]))` (because created and run from the expression), the check flags the variable `fillLength` as not being created before its use.

Yuri05: ok, now I at least understand how it works :)
I have further questions/thoughts regarding this, but I will create separate issues for that.
Accepting the PR for now.

Yuri05 (approved):

PR #343, opened 2022-07-22
Fixes #326 legend title use plot configuration for time profile plots

### LINKED ISSUE #326: Don't include legend title by default
image

This is new, I guess. It should not be in the legend.

pchelle:


IndrajeetPatil:
Thanks, Pierre!

I can confirm that this fixes the legend title issue, but also introduces a new change where the lines now start at 0. Is this expected?

See the differences here:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1037/files

IndrajeetPatil:
I guess the lines used to start from 0 only for the linear scale before, and now the same behaviour is seen for the log scale.

pchelle:
> I guess the lines used to start from 0 only for the linear scale before, and now the same behaviour is seen for the log scale.

@IndrajeetPatil 
Can you circle where the difference is about the lines starting at 0 while they were not before ? (I can't find any difference but the legend title in the commit)

IndrajeetPatil:
That difference turned out to be due to something local. I couldn't reproduce with a fresh build, and so I merged this PR :)

PR #344, opened 2022-07-22
Fixes #333 set cap width of errorbars

### LINKED ISSUE #333: Add caps to error bars
Everywhere they occur, we need to add caps to error bars to be consistent with plots produced by PK-Sim, as specified by @Yuri05 [here](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/945#issuecomment-1185308661). 

Currently, error bars are shown without caps. If we want to retain these defaults, there should at least be a way to add or remove the caps.

image

pchelle:
Currently, width is a global setting applied the same way to all error bars


IndrajeetPatil:
Instead of having this as a global setting, is it possible to make this a field of `tlf::ThemeAestheticSelections`?

This way, I can do something like the following:

```r
  errorbarsConfiguration <- tlf::ThemeAestheticSelections$new(
    size = generalPlotConfiguration$errorbarsSize,
    linetype = generalPlotConfiguration$errorbarsLinetype,
    alpha = generalPlotConfiguration$errorbarsAlpha,
    width = generalPlotConfiguration$errorbarsCapWidth
  )

specificPlotConfiguration$errorbars <- errorbarsConfiguration
```

pchelle:
> ```r
> ThemeAestheticSelections
> ```

It is possible, but it will require more time to make the `ThemeAestheticSelection` dependent on the type of property as cap width is not applicable to ribbons, lines and scatter points.
In the end, this might be for the best since ribbons also behave quite differently (fill property not applicable to any other selection).
It would be cool to link the cap width to the json Theme as well.

IndrajeetPatil:
Okay, in order to not block this PR, I've created a new issue to track this other feature:
https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/347

IndrajeetPatil:
I can't get this to work. 

In https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1038, I am using the new `setDefaultErrorbarCapWidth()` function to set cap width, but it doesn't affect the plots. What am I doing wrong?

pchelle:
@IndrajeetPatil 
Could you try with a different value ?
When testing `ggplot2::geom_errorbar`, it seemed that the `width` argument corresponds to the values of the x-axis and not a size in pts.
In your example, the x-axis ranges from 0 to 250, a width of about 5 or 10 should be visible.

IndrajeetPatil:
I don't think this has anything to do with the range of the data values.

In these plots, the data values are orders of magnitude different, and yet error bar width specification of `0.2` always produces the same width for caps.

``` r
library(tidyverse)

df <- data.frame(
  trt = factor(c(1, 1, 2, 2)),
  resp = c(1, 5, 3, 4),
  group = factor(c(1, 2, 1, 2)),
  upper = c(1.1, 5.3, 3.3, 4.2),
  lower = c(0.8, 4.6, 2.4, 3.6)
)

ggplot(df, aes(trt, resp, colour = group)) + 
  geom_errorbar(aes(ymin = lower, ymax = upper), width = 0.2)
```



``` r

df2 <- df %>%
  mutate(resp = 1e5 * resp, upper = 1e5 * upper, lower = 1e5 * lower)

ggplot(df2, aes(trt, resp, colour = group)) + 
  geom_errorbar(aes(ymin = lower, ymax = upper), width = 0.2)
```



<sup>Created on 2022-07-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

pchelle:
It does but on the values of the **X** axis:
If I use a similar example with x being numeric instead of factor.
You can see the cap width is 0.2 when measuring with the ticks

```r
library(tidyverse)

df <- data.frame(
  trt = c(1, 1, 2, 2),
  resp = c(1, 5, 3, 4),
  group = factor(c(1, 2, 1, 2)),
  upper = c(1.1, 5.3, 3.3, 4.2),
  lower = c(0.8, 4.6, 2.4, 3.6)
)

ggplot(df, aes(trt, resp, colour = group)) + 
  geom_errorbar(aes(ymin = lower, ymax = upper), width = 0.2)
```

image

Then, if I increase the x range
```r
df <- data.frame(
  trt = c(1, 1, 10, 10),
  resp = c(1, 5, 3, 4),
  group = factor(c(1, 2, 1, 2)),
  upper = c(1.1, 5.3, 3.3, 4.2),
  lower = c(0.8, 4.6, 2.4, 3.6)
)

ggplot(df, aes(trt, resp, colour = group)) + 
  geom_errorbar(aes(ymin = lower, ymax = upper), width = 0.2)
```

image

```r
df <- data.frame(
  trt = c(1, 1, 10, 10),
  resp = c(1, 5, 3, 4),
  group = factor(c(1, 2, 1, 2)),
  upper = c(1.1, 5.3, 3.3, 4.2),
  lower = c(0.8, 4.6, 2.4, 3.6)
)

ggplot(df, aes(trt, resp, colour = group)) + 
  geom_errorbar(aes(ymin = lower, ymax = upper), width = 2)
```
image

IndrajeetPatil:
Thanks for the clarification, Pierre!

Hmm, this is indeed a sticky problem then, since we definitely don't want users to compute the exact width of cap needed depending on the range of data values. Typically, users produce such plots in a loop, which means that the cap width argument will have to vary from dataset-to-dataset (i.e., for each iteration).

@Yuri05 and @PavelBal: 

Honestly, given these complications, I'd much rather just leave the caps out of the plot. 

Instead of _ospsuite_ plots adapting to PK-Sim plots, I think the PK-Sim plots should adapt to _ospsuite_ plots, and remove the error bar caps. They just add clutter and do not encode any extra information.

IndrajeetPatil:
Another option is to create a helper function in `tlf` that implements some kind of heuristic that creates cap width depending on the range of data values. And then _ospsuite_ can use this helper internally. 

But this is added complexity, and we really want to be sure these caps are worth the added complexity.

msevestre:
Where did this feature request come from?

IndrajeetPatil:
From here: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/945#issuecomment-1185308661

msevestre:
Ok thanks. At any rate, the user should not have to specify a width for the caps. This is purely some internal artefacts IMO

PavelBal:
There are so many entries on the internet showing error bars with caps:

https://stackoverflow.com/questions/66637578/reducing-size-of-error-bar-caps-in-ggplot2
http://www.sthda.com/english/wiki/ggplot2-error-bars-quick-start-guide-r-software-and-data-visualization
https://www.datanovia.com/en/lessons/ggplot-error-bars/

This cannot be something extraordinary.

IndrajeetPatil:
> This cannot be something extraordinary.

It is not. Both my and Pierre's comments show how easy it is to add caps to error bars. 

The issue is not creating them for a single context, the issue is finding a general solution that will produce the same error bar cap width no matter the range of the data.

pchelle:
I may have a better solution:
As suggested by @IndrajeetPatil , we can internally calculate a value that would adapt to the dataset (e.g. mean space between all the different x values) and provide cap width as a ratio of that value.

If I reuse the example above with `tlf` and this solution, I get:
```r
addErrorbar(
x = c(1, 1, 2, 2),
ymin = c(0.8, 4.6, 2.4, 3.6),
ymax = c(1.1, 5.3, 3.3, 4.2),
capWidth = 1
)
```
image
> `capWidth` of 1 would mean the cap fill all the way until the next data

```r
addErrorbar(
x = c(1, 1, 10, 10),
ymin = c(0.8, 4.6, 2.4, 3.6),
ymax = c(1.1, 5.3, 3.3, 4.2),
capWidth = 1
)
```

image

```r
addErrorbar(
x = c(1, 1, 10, 10),
ymin = c(0.8, 4.6, 2.4, 3.6),
ymax = c(1.1, 5.3, 3.3, 4.2),
capWidth = 0.2
)
```

image

Thoughts ?

IndrajeetPatil:
That looks great, @pchelle! Thanks for working on this so quickly.

One remark: We should rename the parameter from `capWidth` to something like `capExtent`.

This is because:

- when the error bar is **horizontal**, the user will have to specify **height**
- when the error bar is **vertical**,  the user will have to specify **width**

msevestre:
>when the error bar is horizontal, the user will have to specify height
when the error bar is vertical, the user will have to specify width

I don't get this. Isn't it something that the user should not have to care about? 

And sorry for the stupid questions: But why  would we ever want to have something like this in the context of our suite?
image


@Yuri05 Have we ever associated an actual meaning to the width of those caps? (or heights). Or is this purely for visual inspections?

IndrajeetPatil:
>  Isn't it something that the user should not have to care about?

Exactly. But for that to happen, the parameter should be renamed to cover both horizontal and vertical error bars.

PavelBal:
> Have we ever associated an actual meaning to the width of those caps? (or heights). Or is this purely for visual inspections?

The width has no meaning. It would be best if the width/height of the caps corresponds to the width/height of the symbols.

pchelle:
> The width has no meaning. It would be best if the width/height of the caps corresponds to the width/height of the symbols.

That is the issue with `ggplot2`, the function adding caps is `geom_errobar` that use an argument based on the data values for displaying the cap extent while it uses an external measure for the symbol size (in pts).

Yuri05:
> The width has no meaning. It would be best if the width/height of the caps corresponds to the width/height of the symbols.

Either this or **fixed** width. IT does not matter if I have 1 or 100 observed data points in the plot - the width (or height :)) should remain the same.

IndrajeetPatil:
@pchelle I don't think this is working as expected as yet. You can see the plots with the current state of the PR [here](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1038/files).

In particular, the height of the cap seems to vary widely depending on the data and the scaling used. Here are two examples:

image

image

Yuri05:
sometimes I really wonder why it's so hard to implement the simplest things with ggplot...

Can we set the caps width to `capExtent/Number_of_plotted_observed_data_points` ? Then it would always have the same fixed size.

pchelle:
> sometimes I really wonder why it's so hard to implement the simplest things with ggplot...

I could not agree more !

pchelle:
I have finally found a cool solution for this: not using `geom_errorbar` at all.
Instead, I will only add the caps as symbols (`"_"` for vertical errorbars and `"|"` for horizontal) in `geom_point`.
So the cap size will correspond to the size of the character in pts (consistent with arguments for lines and symbol sizes)

Yuri05:
> I have finally found a cool solution for this: not using `geom_errorbar` at all. Instead, I will only add the caps as symbols (`"_"` for vertical errorbars and `"|"` for horizontal) in `geom_point`. So the cap size will correspond to the size of the character in pts (consistent with arguments for lines and symbol sizes)

sounds good!

IndrajeetPatil (approved):
Works great now. Thanks!

PR #361, opened 2022-08-10
Fixes #360 add linetype as internal splitting group for simulation range

### LINKED ISSUE #360: `plotPopulationTimeProfile()`: Separate groups by `fill` property
cf. https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1064

pchelle:
@IndrajeetPatil let me know if this fixes the issue with the ospsuite integration
(geom_ribbon is also using `linetype` under the hood as a grouping variable)


IndrajeetPatil (approved):
That works! Thanks.

PR #370, opened 2022-08-16
Fixes #362 introduce qq plots

### LINKED ISSUE #362: Introduce molecule plot `plotQQ`
- `plotQQ` function
- `QQPlotConfiguration`
- `QQDataMapping`

pchelle:


R/plot-qq.R
```diff
@@ -0,0 +1,108 @@
...
+                   plotObject = NULL) {
+  #----- Validation and formatting of input arguments -----
+  if (is.null(data)) {
+    validateIsNumeric(y)
+    data <- data.frame(y = y)
+    dataMapping <- dataMapping %||% QQDataMapping$new(
+      y = ifNotNull(y, "y"),
+      data = data
+    )
+  }
```
Yuri05, line 42: 2 questions on this:
 - what is `y` and when is the function `plotQQ  `called with `data=NULL` and `y != NULL` ?
 - why do we need `ifNotNull `in the line 39? Isn't it assured that y is not null by `validateIsNumeric(y)` before?

pchelle: Using `data` and `dataMapping` are a bit heavy when most of cases are just displaying QQ plots for a vector.
For this reason, there is also an optional `y` input which is transformed internally into a `data` variable and its `dataMapping`

The examples below display the exact same plots
```r
qqData <- data.frame(y = rnorm(100))

plotQQ(y = qqData$y)

plotQQ(data = qqData, dataMapping = QQDataMapping$new(y="y"))
```

image

Yuri05: nice!

Yuri05 (approved):

PR #380, opened 2022-09-09
374 enum listing molecules

pchelle:


R/themes.R
```diff
@@ -339,103 +339,70 @@ ThemeAestheticSelections <- R6::R6Class(
...
-#' @field plotTimeProfile theme properties for `PlotConfiguration` objects as used in function `plotTimeProfile()`
-#' @field plotObsVsPred theme properties for `PlotConfiguration` objects as used in function `plotObsVsPred()`
-#' @field plotBoxWhisker theme properties for `PlotConfiguration` objects as used in function `plotBoxWhisker()`
-#' @field plotTornado theme properties for `PlotConfiguration` objects as used in function `plotTornado()`
-#' @field plotHistogram theme properties for `PlotConfiguration` objects as used in function `plotHistogram()`
 #' @export
 ThemePlotConfigurations <- R6::R6Class(
   "ThemePlotConfigurations",
+  # This allows the R6 class to accept new fields
+  lock_objects = FALSE,
```
Yuri05, line 346: interesting feature :)

pchelle: Totally.
I think it could be beneficial to apply it to other R6 objects such as workflows or task settings in RE (that would make user defined tasks or settings more flexible)

R/utilities-enums.R
```diff
@@ -1,3 +1,84 @@
+#' @title .getTitlesFromFamilyTag
+#' @description Get all title names from documented functions/R6 classes that have specific family tags.
+#' The function aims at created automated and synchronized enums
+#' @param familyTag Family tag used to document functions or R6 classes
+#' @return character array of titles
+#' @import ospsuite.utils
+#' @keywords internal
+.getTitlesFromFamilyTag <- function(familyTag) {
+  functionNames <- NULL
+  for (filePath in list.files("./R", full.names = TRUE)) {
```
Yuri05, line 10: no idea how it works - but it works :)

R/utilities-enums.R
```diff
@@ -1,3 +1,84 @@
...
+#' @family enum helpers
+Atoms <- enum(c(.getTitlesFromFamilyTag("atom plots")))
+
+#' @title Molecules
+#' @import ospsuite.utils
+#' @export
+#' @description
+#' List of all available molecule plots
+#' @family enum helpers
+Molecules <- enum(c(.getTitlesFromFamilyTag("molecule plots")))
```
Yuri05, line 64: let's rename "Atoms" and "Molecules" to e.g. "AtomPlots" and "MoleculePlots".
Especially "Molecules" has in context of the ospsuite-R completely different meaning and might be confusing here

R/utilities-enums.R
pchelle: When building/documenting a package, the working directory is internally set to the package location during the procedure (I did some testing with `devtools::check()` and `devtools::document()`).

The function `.getTitlesFromFamilyTag` is only called during the build of the package to create the enums.
Thus, when the users check for e.g. `Atoms`, they do not call for `.getTitlesFromFamilyTag` but directly for the resulting list created during the package construction.

Yuri05: ok, cool. Thanks for the explanation :)

Yuri05 (approved):

PR #384, opened 2022-09-23
Fixes #383 histogram can plot bars as frequency

### LINKED ISSUE #383: Histogram option for normalized y axis
s. https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/878

pchelle:


R/histogram-datamapping.R
```diff
@@ -6,6 +6,8 @@ HistogramDataMapping <- R6::R6Class(
   "HistogramDataMapping",
   inherit = XYGDataMapping,
   public = list(
+    #' @field frequency logical defining if histogram displays a frequency in y axis
+    frequency = NULL,
```
Yuri05, line 10: shouldn't it be `frequency = FALSE,` instead of NULL?

R/plot-histogram.R
```diff
@@ -37,6 +41,7 @@ plotHistogram <- function(data = NULL,
                           metaData = NULL,
                           x = NULL,
                           dataMapping = NULL,
+                          frequency = NULL,
```
Yuri05, line 44: same

R/histogram-datamapping.R
pchelle: Here, both work as the actual value is defined during the initialization by `frequency=FALSE`.
Unless `unlock_object=TRUE`, a field named `frequency` has to be available before the initialization of the object no matter its value.

R/plot-histogram.R
pchelle: In plot functions, I had made the choice of leaving `NULL` as the default value for optional arguments.
So that they won't overwrite the values created by `dataMapping` and `plotConfiguration`.
This aims at making it easier to centralize and modify the default `frequency`, `bins`, `distribution`, etc. as it is managed only by the `HistogramDataMapping` object.
I can change it if it makes it clearer for the users.

Yuri05 (approved):

PR #396, opened 2022-12-15
Fixes #392 Create method for dual axis time profile plots

### LINKED ISSUE #392: Right axis methods
Since left and right axes share the same grid in `{ggplot2}`, displaying data for dual axes either require data transformation before display (as illustrated [here](https://r-graph-gallery.com/line-chart-dual-Y-axis-ggplot2.html)) or merging 2 plots in one (as illustrated [here](https://wilkelab.org/cowplot/articles/aligning_plots.html)).
After a few tests, the second solution (merging 2 plots using `{cowplot}` package) seems the most promising

The feature is only required for time profile plots, thus only their input argument should be updated.
Below describes the expected updates:

- `data`/`observedData`: data should include a column with logical values to select which data to be displayed for the right axis (with that implementation, it is easier to display error bars or range plots correctly)
- `metaData`: if the meta data related to that column includes `unit` and `dimension`, it will be used to label the right axis (same principle as for other axes)
- `dataMapping`: a new `y2Axis` argument will be available to map the column selecting the data displayed for right axis
- `plotConfiguration`: will include a `y2Axis` object of class `YAxis` whose field `position` will indicate `"right"`

Internally, the function will create a left and a right plots.
Then, a merging function based on `{cowplot}` functions will align and merge the plots after having removed the background of the right plot.

A dedicated vignette should be created to explain how to include right axis in time profiles

pchelle:


Yuri05:
1 test failed...

Yuri05 (approved):

Issue #413, opened 2023-02-28
Unsopported shape aesthetics


PavelBal:
```
dsLength <- 20
dataSets <- vector("list", dsLength)
 for (i in 1:dsLength){
   ds <- DataSet$new(name = paste0("ds", i))
   ds$setValues(0, 1)
   ds$addMetaData("IndividualId", as.character(i))
   dataSets[[i]] <- ds
 }
myDc <- DataCombined$new()
myDc$addDataSets(dataSets)

plotIndividualTimeProfile(myDc)
```

```
Error in `ggplot2::geom_point()`:
! Problem while converting geom to grob.
ℹ Error occurred in the 6th layer.
Caused by error in `translate_shape_string()`:
! Shape aesthetic contains invalid values: "★", "☆", "♦", "♥", "♠", and "♣"
Run `rlang::last_error()` to see where the error occurred.
```

pchelle:
They are supported, the issue is that they cannot be combined with legal ggplot2 shapes (`"circle"`, `"square"`, etc) because they do not share the same format:  https://stackoverflow.com/questions/67859447/error-when-combining-unicode-shape-with-legal-ggplot2-shapes

For instance, if you run the following example, you get an error
```r
library(ggplot2)

ggplot(
  data = data.frame(x = "", y = c("circle", "pentagon")),
  mapping = aes(x=x,y=y,shape=y)
  ) + 
  geom_point(size = 10) + 
  guides(shape = "none") + 
  scale_shape_manual(values = c("circle", "\u2b1f"))
```
```
Error in `translate_shape_string()`:
! Can't find shape name:
* '⬟'
Run `rlang::last_error()` to see where the error occurred
```

But if circle uses also a character for its shape:

```r
library(ggplot2)

ggplot(
  data = data.frame(x = "", y = c("circle", "pentagon")),
  mapping = aes(x=x,y=y,shape=y)
  ) + 
  geom_point(size = 10) + 
  guides(shape = "none") + 
  scale_shape_manual(values = c("\u25cf", "\u2b1f"))
```

![image](https://user-images.githubusercontent.com/45343665/222434429-c7ca44cc-3215-4925-a431-92670c9e017d.png)

pchelle:
Let me know if PR #414 fixes this

PavelBal:
> Let me know if PR https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/414 fixes this

[...]

PR #414, opened 2023-03-02
Fixes #395 add new shapes

### LINKED ISSUE #395: Add new shapes
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1789

pchelle:
Examples were added to the documentation, which provides a visual display of the linetypes and shapes in the website. For this reason, the doc website was also updated.

Also, note that this fixes a ggplot2 issue that prevents using some shapes together.


PavelBal:
@pchelle Are these symbols actually supported? https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/413

pchelle:
> @pchelle Are these symbols actually supported? #413

Yes. They are supported, the issue is that they cannot be combined with legal ggplot2 shapes ("circle", "square", etc) because they do not share the same format: https://stackoverflow.com/questions/67859447/error-when-combining-unicode-shape-with-legal-ggplot2-shapes

This is exactly the reason why I chose to include only utf characters for drawing all the symbols available in `Shapes`.
I referenced the issue within the documentation of `Shapes` so that it is also available in the website.

Yuri05:
@pchelle Why are the symbols in the new plots so much smaller than before?
grafik

Yuri05:
same here:
https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/414/files?file-filters%5B%5D=.R&file-filters%5B%5D=.json&file-filters%5B%5D=.png&file-filters%5B%5D=.xml&file-filters%5B%5D=.yml&show-viewed-files=false#diff-75802937aa1e49a25f3ec6b5f81f672a834536e8530504d8feba9267ac7717d2
grafik

Yuri05:
>I am updating circle and square, this should be better but might still be slightly smaller than ggplot initial symbols.

@pchelle Can you please regenerate the docu as well?

Yuri05:
@pchelle Smybols are bigger now, but I wonder why e.g. squares became circles now?
grafik

pchelle:
> @pchelle Smybols are bigger now, but I wonder why e.g. squares became circles now? grafik

I have changed the order of symbols in the themes.
Before the shapes were square, **dot**, circle etc but since dot was not useful I changed it to circle, square, diamond, triangle, etc.
That way the default first shape used in most of the plots is a circle. But I can revert to square if necessary.

Yuri05:
no need to revert; I just wanted to be sure that we don't have a square in the enum represented by circle or so :)

Yuri05:
not yet ready for review

docs/dev/reference/Shapes-1.png
pchelle, line 1: This plot (included in the website documentation) displays the current available Shapes

docs/dev/reference/Linetypes-1.png
pchelle, line 1: This plot (included in the website documentation) displays the current available Linetypes

Yuri05 (approved):

Issue #359, opened 2022-08-08
Add legend entries for fold distance lines in `plotObsVsPred()`


IndrajeetPatil:
cf. https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1063

PR #418, opened 2023-03-07
Add legend entries for fold distance lines in `plotObsVsPred()`

Felixmil:
- related to #359
- Draw legend for fold distance lines but not for abline (fold distance = 1)

fold-distance-several-folds

### DIFF
--- a/R/utilities-molecule-plots.R
+++ b/R/utilities-molecule-plots.R
@@ -226,8 +226,14 @@
     "diagonal" = ggplot2::geom_abline(
       slope = 1,
       intercept = value,
-      color = aestheticValues$color,
       linetype = aestheticValues$linetype,
+      color = aestheticValues$color,
+      alpha = aestheticValues$alpha,
+      size = aestheticValues$size
+    ),
+    "obsvspredDiagonal" = ggplot2::geom_abline(
+      aes_(slope = 1, intercept = value, linetype = as.character(position)),
+      color = aestheticValues$color,
       alpha = aestheticValues$alpha,
       size = aestheticValues$size
     ),
--- a/R/utilities-molecule-plots.R
+++ b/R/utilities-molecule-plots.R
@@ -235,7 +235,8 @@
       aes_(slope = 1, intercept = value, linetype = as.character(position)),
       color = aestheticValues$color,
       alpha = aestheticValues$alpha,
-      size = aestheticValues$size
+      size = aestheticValues$size,
+      key_glyph = plotObject$plotConfiguration$getKeyGlyph()
     ),
     "ddiHorizontal" = ggplot2::geom_abline(
       slope = 0,
@@ -255,13 +256,13 @@
 #' @param rightPlotObject A `ggplot` object with right y-axis
 #' @return A `ggplot` object with dual y-axis
 #' @export
-getDualAxisPlot <- function(leftPlotObject, rightPlotObject){
+getDualAxisPlot <- function(leftPlotObject, rightPlotObject) {
   stopifnot(requireNamespace("cowplot", quietly = TRUE))
   # Only one legend shall be kept to prevent text not aligned and on top of plot axes text
   # For most cases, right plot legend is kept as is while left plot legend is removed
   # If left side legend, left plot legend is kept as is while right plot legend is removed
   legendPosition <- getLegendPosition(leftPlotObject)
-  if(isIncluded(legendPosition, LegendPositions$outsideLeft)){
+  if (isIncluded(legendPosition, LegendPositions$outsideLeft)) {
     rightPlotObject <- setLegendPosition(rightPlotObject, position = LegendPositions$none)
   } else {
     leftPlotObject <- setLegendPosition(leftPlotObject, position = LegendPositions$none)
@@ -270,12 +271,12 @@ getDualAxisPlot <- function(leftPlotObject, rightPlotObject){
   leftScale <- ggplot2::layer_scales(leftPlotObject)
   rightScale <- ggplot2::layer_scales(rightPlotObject)
   mergeXRange <- range(leftScale$x$range$range, rightScale$x$range$range)
-  
+
   leftPlotObject <- setXAxis(leftPlotObject, limits = mergeXRange)
   rightPlotObject <- setXAxis(rightPlotObject, limits = mergeXRange)
-  
+
   # Transformed right plot to be compatible with left plot
-  rightPlotObject <- rightPlotObject + 
+  rightPlotObject <- rightPlotObject +
     ggplot2::theme(
       # Update right axis properties
       axis.text.y.right = rightPlotObject$plotConfiguration$y2Axis$font$createPlotFont(),
@@ -284,8 +285,8 @@ getDualAxisPlot <- function(leftPlotObject, rightPlotObject){
       panel.grid.major.y = rightPlotObject$plotConfiguration$background$y2Grid$createPlotElement(),
       panel.grid.minor.y = rightPlotObject$plotConfiguration$background$y2Grid$createPlotElement(
         size = as.numeric(rightPlotObject$plotConfiguration$background$y2Grid$size) / 2
-        ),
-      # Remove all other background properties 
+      ),
+      # Remove all other background properties
       plot.background = ggplot2::element_blank(),
       panel.background = ggplot2::element_blank(),
       axis.line.x = ggplot2::element_blank(),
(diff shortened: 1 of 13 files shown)


Yuri05:
Questions/Remarks: 
1. Please add an (optional) parameter which turns the legend for the fold lines on and off (Defalt in my opinion should be OFF, but we can discuss)
2. Why is the legend implemented for **pred-vs-observed** plots but not for **residuals-vs-pred plots?**
3. The default legend position **top-right** is not optimal for this kind of plots - e.g. **top-left** or **bottom-right** would be much better
4. Why using **diagonal** lines for the legend? Horizontal lines are better in my opinion, but we can discuss.

Felixmil:
> Questions/Remarks:
> 
> 1. Please add an (optional) argument which turns the legend for the fold lines on and off (Default in my opinion should be OFF, but we can discuss)
>    EDIT: actually this argument should be part of the `plotConfiguration ` passed to `plotObsVsPred `
> 2. Why is the legend implemented for **pred-vs-observed** plots but not for **residuals-vs-pred plots?**
> 3. The default legend position **top-right** is not optimal for this kind of plots - e.g. **top-left** or **bottom-right** would be much better
> 4. Why using **diagonal** lines for the legend? Horizontal lines are better in my opinion, but we can discuss (this could become a setting in the `plotConfiguration  `argument as well)

Many good points.

1. Ok, will do. This could also be done in the `ObsVsPredPlotConfiguration` subclass, what do you think of that ?
2. The issue stated `plotObsVsPred()`, I wanted an initial review before implementing this feature elsewhere but will do in this PR. Do you think any other plotting function should have this ? (I saw that a `plotDDIRatio()` also has a `foldDistance` argument)
3. Ok, this also could be set in `ObsVsPredPlotConfiguration` subclass right ?
4. Diagonal lines are the default legend for `ggplot2::geom_abline()` i'll look for a way to change this and make this an option. Horizontal line by default will be better indeed.

Yuri05:
>     1. Ok, will do. This could also be done in the `ObsVsPredPlotConfiguration` subclass, what do you think of that ?

Yes!
>     2. The issue stated `plotObsVsPred()`, I wanted an initial review before implementing this feature elsewhere but will do in this PR. Do you think any other plotting function should have this ? (I saw that a `plotDDIRatio()` also has a `foldDistance` argument)

I don't mind if other plot types are implemented in other PRs - just wanted to make sure it was not forgotten :)

Regarding DDIRatio plots (and PKRatio plots): was not a requirement on our side so far.
@sfrechen @twendl @incei @andredlm @PavelBal Do you have any opinion on this? Should we provide the possibility to display the fold distance legend for DDI/PK-Ratio plots?

At the moment, the description of the fold lines for those plots is included into the **figure caption**, e.g.
https://github.com/Yuri05/Iba_Pediatric_Qualification_Package_P-gp_Ontogeny/blob/review/report/report.md#figure-2-1
**My** opinion on this: the current way of displaying the fold distance info is sufficient and **we don't need to display it in the plot legend**.

>     3. Ok, this also could be set in `ObsVsPredPlotConfiguration` subclass right ?

Yes. And I would suggest that the **default** legend location is already e.g. top-left (because top-right of ObsVsPred is always covered by the plot). Or maybe top-center-**outside**

>     4. Diagonal lines are the default legend for `ggplot2::geom_abline()` i'll look for a way to change this and make this an option. Horizontal line by default will be better indeed.

👍

PavelBal:
> At the moment, the description of the fold lines for those plots is included into the figure caption, e.g.
https://github.com/Yuri05/Iba_Pediatric_Qualification_Package_P-gp_Ontogeny/blob/review/report/report.md#figure-2-1
My opinion on this: the current way of displaying the fold distance info is sufficient and we don't need to display it in the plot legend.

No strong opinion, can be left out for now.

Felixmil:
Here is what you can now get:

```r
obsVsPredData <- data.frame(x = c(1, 2, 1, 2, 3), y = c(5, 0.2, 2, 3, 4))

plotConfiguration <- ObsVsPredPlotConfiguration$new(
    foldLinesLegend = TRUE,
    xScale = Scaling$log, xLimits = c(0.05, 50),
    yScale = Scaling$log, yLimits = c(0.05, 50),
)

plotObsVsPred(
      data = obsVsPredData,
      dataMapping = ObsVsPredDataMapping$new(x = "x", y = "y"),
      plotConfiguration = plotConfiguration,
      foldDistance = c(1, 5, 10, 15, 20, 25)
    )
```
fold-distance-many-folds

```r
plotObsVsPred(
      data = obsVsPredData,
      dataMapping = ObsVsPredDataMapping$new(x = "x", y = "y"),
      plotConfiguration = ObsVsPredPlotConfiguration$new(
        foldLinesLegend = TRUE,
        foldLinesLegendDiagonal = TRUE,
        legendTitle =  "diagonal lines",
        xScale = Scaling$log, xLimits = c(0.05, 50),
        yScale = Scaling$log, yLimits = c(0.05, 50)),
      foldDistance = c(1, 5, 10, 15, 20, 25)
 )
```

fold-distance-many-folds-diagonal

PavelBal:
When creating a observed-vs-simuated from ospsuite-r, the legend for the first fold line (in this case, 2-fold) looks like a solid line. Maybe just some legend settings?:

image



```
# load the simulation
sim <- loadTestSimulation("MinimalModel")
simResults <- importResultsFromCSV(
  simulation = sim,
  filePaths = getTestDataFilePath("Stevens_2012_placebo_indiv_results.csv")
)

# import observed data (will return a list of `DataSet` objects)
dataSet <- loadDataSetsFromExcel(
  xlsFilePath = getTestDataFilePath("CompiledDataSetStevens2012.xlsx"),
  importerConfiguration = loadDataImporterConfiguration(getTestDataFilePath("ImporterConfiguration.xml"))
)

# create a new instance and add datasets
myCombDat <- DataCombined$new()
myCombDat$addDataSets(dataSet)
myCombDat$addSimulationResults(
  simResults,
  quantitiesOrPaths = c(
    "Organism|Lumen|Stomach|Metformin|Gastric retention",
    "Organism|Lumen|Stomach|Metformin|Gastric retention distal",
    "Organism|Lumen|Stomach|Metformin|Gastric retention proximal"
  )
)

myCombDat$setGroups(
  names = c(
    "Organism|Lumen|Stomach|Metformin|Gastric retention",
    "Organism|Lumen|Stomach|Metformin|Gastric retention distal",
    "Organism|Lumen|Stomach|Metformin|Gastric retention proximal",
    "Stevens_2012_placebo.Placebo_total",
    "Stevens_2012_placebo.Placebo_distal",
    "Stevens_2012_placebo.Placebo_proximal"
  ),
  groups = c("Solid total", "Solid distal", "Solid proximal", "Solid total", "Solid distal", "Solid proximal")
)
plotObservedVsSimulated(myCombDat, foldDistance = c(2, 3))
```

Yuri05:
> Here is the result with the latest commit:

nice!

Felixmil:
> > Here is the result with the latest commit:
> 
> just 1 question: what are the vertical lines over the triangles? grafik
> 
> It's not the SD I assume (otherwise it would be a horizontal line).


I think its the vertical limits of the SD's

Yuri05:
> I think its the vertical limits of very low (null?) SD's

Yes, probably that's it.
If SD=null, it should not be displayed at all. If it's very low - well, than it's probably OK to display it like now.
Anyways, it's not related to the current PR :)

R/utilities-molecule-plots.R
```diff
@@ -226,11 +226,18 @@
...
       slope = 1,
       intercept = value,
-      color = aestheticValues$color,
       linetype = aestheticValues$linetype,
+      color = aestheticValues$color,
       alpha = aestheticValues$alpha,
       size = aestheticValues$size
     ),
+    "obsvspredDiagonal" = ggplot2::geom_abline(
+      aes_(slope = 1, intercept = value, linetype = as.character(position)),
```
Yuri05, line 235: @Felixmil 
I don't understand this: `linetype = as.character(position)` Could you explain?

Also "obsvspredDiagonal" must be included into the function header.
```
#' @param plotObject A `ggplot` object
#' @param type one of "horizontal", "vertical" or "diagonal"
#' Note that for "diagonal", geom_abline is used.
```

Felixmil: - position is an incremental integer that is generated from the `lineIndex` id in the [for loop that draw each line one after another](https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/a9050fd46bcabb2cb9af4953e96a965a07e58837/R/plot-obs-vs-pred.R#L82-L92) . So each abline set has an ID,
- To make ggplot draw each abline with different linetype, this ID is included in the `aes()`,
- However, ggplot throws a warning when passing numeric variable to `linetype`, as it recommends categorical variable.

➡️ **ID is transformed to character to match ggplot recommendations.**

*Note that instead of character, we could transform `position` to a factor.*

Yuri05: somehow for me it works completely different (ggplot2 version 3.4.1): passing the line type as integer works fine, **passing line type as character crashes**. 
```
> df <- data.frame(x = 1:5, y = 1:5)
> ggplot(df, aes(x, y)) + geom_abline(slope=1, intercept=0.1, linetype = 2)
> ggplot(df, aes(x, y)) + geom_abline(slope=1, intercept=0.1, linetype = "dashed")
> ggplot(df, aes(x, y)) + geom_abline(slope=1, intercept=0.1, linetype = "3313")
> ggplot(df, aes(x, y)) + geom_abline(slope=1, intercept=0.1, linetype = as.character(2))
Error in grid.Call.graphics(C_segments, x$x0, x$y0, x$x1, x$y1, x$arrow) : 
  invalid line type: must be length 2, 4, 6 or 8
```

According to the online help, strings are only accepted if they represent "an even number of hexadecimal digits":
grafik

Confused...

Yuri05: Another question: `aes_` (with underscore) seems to be a deprecated function and `aes `should be used instead?
grafik

Felixmil: > ok, it works as implemented using `geom_abline(aes(...))` instead of `geom_abline(...)`. Weird ggplot.
> 
> obsolete comment

Yes because when you set it outside `aes()` the integer is used as an alias to a linetype i.e dashed, dotted etc... But when it is supplied to `aes()` you don't give a linetype but a group identifier, then the `scale_linetype` will map these identifiers to actual linetypes.

> Another question: `aes_` (with underscore) seems to be a deprecated function and `aes `should be used instead?

`aes_string()` is the same family of `aes_()` and is also depreciated. However, it is used absolutely everywhere in the package. The new way of passing arguments to plot generating functions are described [here](https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html#using-aes-and-vars-in-a-package-function). I am currently testing this and how to make the package compatible with this new paradigm in  #422.

Yuri05: ok, thx for the explanation :)
weird ggplot

Issue #423, opened 2023-03-13
Replace all `aes_string()` calls by `aes()`


Felixmil:
`aes_string()` is deprecated, and [this](https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html#using-aes-and-vars-in-a-package-function) should be used.

However, while the `dataMap()` method adds backticks <code>`</code> to secure column names, ggplot2 also does it when using the new syntax. 

So changing the `aes_string()` calls would also need to update the `dataMap` methods.

PR #424, opened 2023-03-13
Update all deprecated `aes_string()` call to `aes()` with .data pronoun

Felixmil:
Followed ggplot2 recommendation for inclusion in this package. 
As stated in #423 `aes_string()` is deprecated and using `aes()` will allow us to do more things in a simpler way.

> **Note**
> All tests passed on `{osps}` after installing this branch as local version. However, since this change impact basically all the plotting function, it may be considered as a breaking change.


R/plot-histogram.R
```diff
@@ -109,21 +109,33 @@ plotHistogram <- function(data = NULL,
...
+      fill = .data[[mapLabels$fill]]
+    )
     if (dataMapping$stack) {
       # If histogram bars are stacked,
       # Calculate overall frequency as count per bin / total
       # This results in same histogram shapes no matter the data groups
-      yAes <- "..count../sum(..count..)"
+      mapping <- ggplot2::aes(
+        x = .data[[mapLabels$x]],
+        y = after_stat(coun / sum(count)),
```
svavil, line 136: Is `coun` a mistype here?

svavil:
Is it expected that many plot markers change from squares to circles? Why is that?

Felixmil:
> Is it expected that many plot markers change from squares to circles? Why is that?

This is comming from #414, please pull latest version of the branch.

svavil:
> > Is it expected that many plot markers change from squares to circles? Why is that?
> 
> This is comming from #414, please pull latest version of the branch.

I don't mean that the markers change in my local testing, but rather than #424 includes these changes in the test png files.

Felixmil:
> > > Is it expected that many plot markers change from squares to circles? Why is that?
> > 
> > 
> > This is comming from #414, please pull latest version of the branch.
> 
> I don't mean that the markers change in my local testing, but rather than #424 includes these changes in the test png files.

- #414  was merged into develop while #418 was waiting for merge. #414 changed shape level 1 from square to points which made the #414 test png outdated,
- #418 was merged without generating new test png,
- `develop`, containing outdated pngs was merged into this branch,
- Test run in this branch updated the pngs which were committed in order to make the build pass on appveyor.

PavelBal:
Fine with me, but I think @pchelle should review it first.

PavelBal:
@pchelle Please review at your earliest convenience :)

R/plot-histogram.R
Felixmil: I think using {lintr} would be great. There's already a Github action ready to go [here](https://github.com/r-lib/actions/tree/old/examples#lint-workflow).

We would need to customize it a little and define which [linter](https://lintr.r-lib.org/reference/linters.html) we want to apply. But that depends on the scope of the verification we want to implement.

Maybe we can start with the most basic level: make sure that the code can be evaluated. I can make some test if required.

PavelBal (approved):

PavelBal:
@Yuri05 I am going to merge today/tomorrow, if no one objects.

Yuri05:
@PavelBal I would like to have the approval by @pchelle before merging

PavelBal:
> I would like to have the approval by @pchelle before merging

Sure, let's wait then.

msevestre:
@pchelle Please review asap. Thanks

pchelle (approved):
All good.
`after_stat` was introduced in ggplot2 starting version 3.3.0, so there is no compatibility issue.
should we declare `ggplot2::after_stat` instead to stay consistent with all the other calls to ggplot2 ?

Felixmil:
> should we declare `ggplot2::after_stat` instead to stay consistent with all the other calls to ggplot2 ?

Yes. Done in latest commit.

PR #430, opened 2023-03-23
New Dependency / Installation proposal

Felixmil:
## Easier package installation
I also updated the README and DESCRIPTION file to make the package installation/dependencies management easier.

Users can install latest {tlf} **released** version using:
```r
# install.packages("remotes")
remotes::install_github("Open-Systems-Pharmacology/TLF-Library@*release")
```

Now that the `DESCRIPTION` file contains the following line:
```
Remotes: Open-Systems-Pharmacology/OSPSuite.RUtils@*release
```
It will directly get the latest **released** version of `{ospsuite.utils}` without any manual installation from the user.


Felixmil:
The build is failing because github api is limiting the rate of request and then fail to install dependency. I managed to make this work by adding a `GITHUB_PAT` environment variable using the Appveyor web interface in a [private appveyor project](https://ci.appveyor.com/project/Felixmil/tlf-library/builds/46585709).

@Yuri05  Do you have access to the appveyor's build configuration for this repo?

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: tlf
 Type: Package
 Title: TLF Library
-Version: 1.5.0
+Version: 1.5.125.9000
```
PavelBal, line 4: Hmmm we don't do this, the minor version should be set by AppVeyor.

DESCRIPTION
```diff
@@ -114,3 +114,4 @@ Collate:
     'utilities-molecule-plots.R'
     'utilities-theme.R'
     'utils.R'
+Remotes: github::Open-Systems-Pharmacology/OSPSuite.RUtils
```
Yuri05, line 117: hmm, I would not do this as well. 
It would install the **development** version of ospsuite.utils, which in the most cases is NOT what the user wants 
(user usually needs to install a **released** version instead)

DESCRIPTION
Yuri05: End users should always installe **released** versions anyways (those from https://github.com/Open-Systems-Pharmacology/TLF-Library/releases).
And if for some reason you need to install a dev version: I think, you would know what you have installed?
And if for some reason you don't know it anymore - just check if the version of the installed package is in the list of the released versions (https://github.com/Open-Systems-Pharmacology/TLF-Library/tags)

Some time ago we indeed used the X.Y.Z.9000 versioning - it's more headache than with the current approach.

DESCRIPTION
Felixmil: The package release mechanics would be the following:
- in `develop` branch so ongoing dev, have `Remotes::user/pkg` that target dev version of dependencies.
- For a release, this same line is changed to  `Remotes::user/pkg@*release` to target the latest release available on github.
- Then the code is changed back to `Remotes::user/pkg` to pursue dev.

This way:
1. When a user install from a release he always get the **released** dependencies.
2. When a dev installs a dev version, he gets the dev version of dependencies.

As the different OSP packages are heavily depending on each others and are developed in tandem, making point 2 automatic is significantly safer. 
This also allow the user to install the packages in a easier way without needing to maintain links to release on our side. (btw the link for installing {ospsuite.utils} is out of date in the package website).

Together with this, ones usually:
A. Update the NEWS file
   - For release: Increment the NEWS section with the version number,
   - After release: Add a new header `# pkg (development version)

B. Update the version number
  - For release: match the version with what was provided by AppVeyor and remove the `.9000`,
  - After release: add the `.9000` again.

See these two commits for a real example:
- https://github.com/tidyverse/tidyr/commit/561fb8eeab6eb229f84646a1123890775846c978
- https://github.com/tidyverse/tidyr/commit/d4b730a858b4bbdd3ffd4250250e982b160fddc8

README.md
```diff
@@ -18,31 +18,23 @@ The `{tlf}` package provides an object-oriented framework to create
...
 ## Installation
 
-You can install the development version of `{tlf}` from
-[GitHub](https://github.com/) with:
-
+You can install the latest release of `{tlf}` directly from Github with:
 ``` r
 # install.packages("remotes")
-remotes::install_github("Open-Systems-Pharmacology/TLF-Library")
+remotes::install_github("Open-Systems-Pharmacology/TLF-Library@*release")
```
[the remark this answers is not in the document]
Felixmil: Because together with this line in DESCRIPTION
```
Remotes: github::Open-Systems-Pharmacology/OSPSuite.RUtils@*release
```

It will install the packages and its **releases** dependencies in just one line of code. Easier, no ?

DESCRIPTION
Yuri05: >Do we have specific rules regarding when releases are made ? 

tlf and utils are usually released when it's required by the new release of either ospsuite-R or ospsuite.reportingengine (RE).
ospsuite-R is usually released with the new OSP release plus on-demand releases in between (e.g. when important new features are available) or if it's required by the new release of RE.
RE is released when all features planned for the next release are implemented.

>But they should, right ? At least within OSP universe.

Not nesesserily. E.g. if we have a release of RE, which relies on the tlf V2; then we create tlf release V3 which is not compatible with V2 but there is no need for the new RE release. Then the latest RE does not work with the latest tlf.

appveyor.yml
```diff
@@ -52,7 +54,7 @@ build_script:
   - travis-tool.sh install_deps
   - travis-tool.sh r_binary_install curl
   - Rscript -e "install.packages(c('covr', 'spelling'), repos = 'http://cran.us.r-project.org', type='win.binary')"
-  - Rscript -e "download.file('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/ospsuite-rutils/artifacts/ospsuite.utils.zip?pr=false', destfile = 'ospsuite.utils.zip', mode='wb'); install.packages('ospsuite.utils.zip', repos = NULL, type = 'binary')"
+  # - Rscript -e "download.file('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/ospsuite-rutils/artifacts/ospsuite.utils.zip?pr=false', destfile = 'ospsuite.utils.zip', mode='wb'); install.packages('ospsuite.utils.zip', repos = NULL, type = 'binary')"
```
PavelBal, line 57: Not required because of `Remotes: github::Open-Systems-Pharmacology/OSPSuite.RUtils@*release` in the `DESCRIPTION`?

PavelBal:
@Felixmil Is this still of relevance?

Felixmil:
> @Felixmil Is this still of relevance?

I would say yes in order to make package installation easier. But the discussions led to the conclusion that we do not want to change the way versionning is handled in ospsuite packages.

We kinda implemented this in esqlabsR so we can refer to it if we want to adopt the same installation/versionning process for tlf and other package so let's close this.

PR #448, opened 2023-03-30
447 display lloq in vs plots

Felixmil:
Example:
image


PavelBal:
I think only the vertical line should be shown, as the simulated results do not really have an LLOQ.

PavelBal:
> Is Observed always plotted on x ?

At least for `plotSimulatedVsObserved()` from `ospsuite`. I don't know if someone would use the `TLF` functions directly to do it the other way around. So theoretically it could be vice versa. Can't the orientation of the LLOQ line be dependent on the axis of observed data? So basically it should only be shown for observed data, not for simulation results.

In theory even YES, we could plot observed vs observed, in this case LLOQ lines would be on both axis...

Felixmil:
> Can't the orientation of the LLOQ line be dependent on the axis of observed data?

Well `ObsVsPredDataMapping` does not make the difference between obs and pred, just x and y. 

> In theory even YES, we could plot observed vs observed, in this case LLOQ lines would be on both axis...

True. But this would not be the intented use case of `plotObsVsPred` function then. However, I agree that the function should have enough flexibility to be used in different context while having appropriate and expected default behavior.

# Proposed solution
Add a `direction` setting in the  `ObsVsPredConfiguration` with default to "vertical".

## default plotConfiguration
```r
plotObsVsPred(
      data = obsVsPredData,
      dataMapping = ObsVsPredDataMapping$new(x = "x", y = "y",
                                             lloq = "lloq")
    )
```
image

## Horizontal LLOQ (predVsObs)

```r
plotObsVsPred(
      data = obsVsPredData,
      dataMapping = ObsVsPredDataMapping$new(x = "x", y = "y",
                                             lloq = "lloq"),
      plotConfiguration = ObsVsPredPlotConfiguration$new(lloqDirection = "horizontal")
    )
```

image

## Both

```r
plotObsVsPred(
      data = obsVsPredData,
      dataMapping = ObsVsPredDataMapping$new(x = "x", y = "y",
                                             lloq = "lloq"),
      plotConfiguration = ObsVsPredPlotConfiguration$new(lloqDirection = "both")
    )
```

image

## Both + Groups

image

PR #452, opened 2023-04-04
Fix typos in vignettes and other documentation

Felixmil:


inst/WORDLIST
```diff
@@ -1,34 +1,109 @@
...
+DataMapping
+DataMappings
+DefaultDataMappingValues
+Enum
+ExportConfiguration
+ExportFormats
+ExportUnits
+FontFaces
+GPLv
+Github
```
Yuri05, line 29: The correct spelling is GitHub (with capital H)...

inst/WORDLIST
```diff
@@ -37,25 +112,17 @@ createPlotTicks
...
 errorbar
 errorbars
 et
-ExportConfiguration
-ExportFormats
 exportPlot
 exportPlotConfigurationCode
-ExportUnits
-FontFaces
+fonctions
```
Yuri05, line 124: that's a typo which should be fixed rather than entered into the word list

Felixmil: French bias here. Fixed in next commit.

PR #458, opened 2023-04-11
make foldLineslegendType active

Felixmil:


PavelBal:
@Felixmil Could you briefly describe why this is required?

Felixmil:
> @Felixmil Could you briefly describe why this is required?

`foldLineslegendType` was "updated" depending on `foldLineslegendDiagonal` **only during object initialization**. If ones would change the `foldLineslegendDiagonal` setting to `TRUE` after initializing it with `FALSE` , `foldLineslegendType`'s value would still be `path` instead of `abline`.

Now, as it is as active field, a fonction is called when `foldLineslegendType` is evaluated. this function checks the **current** value of  `foldLineslegendDiagonal` and returns the proper geom for the legend key.

PavelBal (approved):

PR #459, opened 2023-04-11
Fixes #455 tlf default properties for obs vs pred

### LINKED ISSUE #455: Default properties for obs vs pred plots
Default legend position should actually be set in `{ospsuite}` package since the `{tlf}` default property is used by `{ospsuite.reportingengine}` leading to inconsistent legend positions across the reported figures.

pchelle:
Default properties to be defined at ospsuite level to prevent issues with  reporting engine properties relying on tlf default


Yuri05:
@pchelle Some snapshot tests fail

pchelle:
@Felixmil 
Just to let you know if/when this is merged, the behaviour for `{ospsuite}` will change for obs vs pred plots:
- you will need set the default legend position there
- identity line will be displayed by default (but since the `foldDistance` parameter may be used by default within `{ospsuite}`, this may not have much impact unless the parameter ends up `NULL`)

Felixmil:
> @Felixmil Just to let you know if/when this is merged, the behaviour for `{ospsuite}` will change for obs vs pred plots:
> 
> * you will need set the default legend position there
> * identity line will be displayed by default (but since the `foldDistance` parameter may be used by default within `{ospsuite}`, this may not have much impact unless the parameter ends up `NULL`)

Alright, I already ported some new TLF features' settings [there](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1239), I will add the legend position.

Thanks for letting me know.

PR #461, opened 2023-04-13
Fixes #460 viridis color palettes

### LINKED ISSUE #460: Color palettes can use viridis

pchelle:


R/plot-cumulative-time-profile.R
```diff
@@ -99,15 +99,25 @@ plotCumulativeTimeProfile <- function(data = NULL,
...
   if (!isEmpty(plotConfiguration$colorPalette)) {
-    try(suppressMessages(
-      plotObject <- plotObject +
-        ggplot2::scale_fill_brewer(
-          palette = plotConfiguration$colorPalette,
-          aesthetics = c("color", "fill")
-        )
-    ))
+    if(!isIncluded(plotConfiguration$colorPalette, .ViridisPalettes)){
+      try(suppressMessages(
```
PavelBal, line 103: What happens when `try` fails?

pchelle: You get only a warning message:

```
Warning message:
In pal_name(palette, type) : Unknown palette xxx
```

pchelle: `suppressMessages` is only intended to prevent the ggplot annoying message stating that there was already a scale on colors

PavelBal: But then no palette is set? I am just confused that there is no `catch` for the `try`.

R/plot-cumulative-time-profile.R
```diff
@@ -99,15 +99,25 @@ plotCumulativeTimeProfile <- function(data = NULL,
...
+        plotObject <- plotObject +
+          ggplot2::scale_fill_brewer(
+            palette = plotConfiguration$colorPalette,
+            aesthetics = c("color", "fill")
+          )
+      ))
+    }
+    if(isIncluded(plotConfiguration$colorPalette, .ViridisPalettes)){
```
Yuri05, line 111: why not just `else `instead of `if(isIncluded`

R/plot-tornado.R
```diff
@@ -159,13 +159,24 @@ plotTornado <- function(data = NULL,
...
+    if(!isIncluded(plotConfiguration$colorPalette, .ViridisPalettes)){
+      try(suppressMessages(
+        plotObject <- plotObject +
+          ggplot2::scale_fill_brewer(
+            palette = plotConfiguration$colorPalette,
+            aesthetics = c("color", "fill")
+          )
+      ))
+    }
+    if(isIncluded(plotConfiguration$colorPalette, .ViridisPalettes)){
```
Yuri05, line 171: same: why not just `else `instead of `if(isIncluded`

R/plot-cumulative-time-profile.R
Yuri05: also I would exchange the If and the else part (just to avoid the negation in the if):
```
if(isIncluded(plotConfiguration$colorPalette, .ViridisPalettes)){
...
else {
...
}
```

R/plot-cumulative-time-profile.R
pchelle: Indeed, then I have set a `tryCatch` instead. Here is corresponding outcome from the simple example below:
```r
plotTornado(
    x = rnorm(5),
    y = letters[1:5],
    colorPalette = "xxx"
)
```
```
Warning message:
Unknown palette 'xxx', using palette 'Set1' instead.
See enum 'ColorPalettes' to check available color palettes.
```
image

R/plot-cumulative-time-profile.R
```diff
@@ -99,15 +99,43 @@ plotCumulativeTimeProfile <- function(data = NULL,
...
+            plotObject <- plotObject +
+              ggplot2::scale_fill_brewer(
+                palette = plotConfiguration$colorPalette,
+                aesthetics = c("color", "fill")
+              )
+          )
+        }
+      )
+    }
   }
```
Yuri05, line 138: @pchelle 
it seems that the code here and in plot-tornado is the same.
Can you extract it into a function (e.g. `applyColorPaletteTo(plotObject, plotConfig) `)?

Yuri05 (approved):

PR #466, opened 2023-04-27
Values and axis limits

Felixmil:
We can now decide to set limits on values or/and axis. see [these examples](https://ggplot2.tidyverse.org/reference/coord_cartesian.html#ref-examples) for explanations.

Setting limits on values can help reducing time for drawing plot while axis limits act just as a zoom on the plot.


R/plotconfiguration-axis.R
```diff
@@ -314,6 +334,7 @@ XAxisConfiguration <- R6::R6Class(
       suppressMessages(
         plotObject <- plotObject +
           ggplot2::scale_x_continuous(
```
Felixmil, line 336: `scale_` filters the data

pchelle (commented):
@Felixmil 
Would it be okay if we keep and deprecate `limits` ? (linking the deprecated `limits` to `axisLimits`)
I am worried that some/many functions of the reporting engine are relying on this property and might break from this PR

Felixmil:
> Would it be okay if we keep and deprecate `limits` ? (linking the deprecated `limits` to `axisLimits`)
> I am worried that some/many functions of the reporting engine are relying on this property and might break from this PR

Yes sure, should a message or warning appear to tell the user to use `axisLimits` instead of `limits` ?

pchelle:
> > Would it be okay if we keep and deprecate `limits` ? (linking the deprecated `limits` to `axisLimits`)
> > I am worried that some/many functions of the reporting engine are relying on this property and might break from this PR
> 
> Yes sure, should a message or warning appear to tell the user to use `axisLimits` instead of `limits` ?

I think it may be more appropriate to throw a warning indeed. I'll create an RE issue to handle such warnings.

Felixmil:
@pchelle Done in last commits

pchelle (approved):

PR #478, opened 2023-08-16
Fixes #477 better handling of non available angles

### LINKED ISSUE #477: Wrong warning text
grafik

I assume it should be 180 and not 189? Also the value at the end seems to be missing

pchelle:


R/plotconfiguration-label.R
```diff
@@ -54,16 +54,12 @@ LabelConfiguration <- R6::R6Class(
...
         if(!isOfType(label, "Label")) {
-          labels[[labelName]] <- asLabel(text = label, font = currentTheme$fonts[[labelName]])
+          label <- asLabel(text = label, font = currentTheme$fonts[[labelName]])
         }
-
-        # Check Chosen angle is available
-        availableAngles <- c(0,90,180,270)
-        if (!(labels[[labelName]]$font$angle %in% availableAngles)) {
-          labels[[labelName]]$font$angle <- availableAngles[which(abs(label$font$angle - availableAngles) == min(abs(label$font$angle - availableAngles)))][1]
-          warning("Angles other than 0, 90, 189 and 270 are not available for title, subtitles, caption and axis titles. Replacing by closest available value: ", label$font$angle,".")
```
pchelle, line 64: Since label is `NULL` at the beginning for most cases and was not updated by the if loop, `label$font$angle` was still `NULL` in the warning message.

R/plotconfiguration-label.R
```diff
@@ -54,16 +54,12 @@ LabelConfiguration <- R6::R6Class(
...
-        }
-        private[[paste0(".", labelName)]] <- asLabel(labels[[labelName]])
+        # Check chosen angle is available
+        # use modulo 360 to in case minus angles were provided
+        label$font$angle <- .checkIsInAvailableAngles(label$font$angle %% 360)
```
pchelle, line 61: To make it more robust, I have used the modulo function: `-90 %% 360 = 270`

pchelle:
@Yuri05 
I need to check if this fixes the warning message in the Reporting Engine.
Btw, the initial tlf update only affects the titles. Thus, RE boxplots should still have their simulation set names with a 45 $\degree$ angle

Felixmil (approved):

PR #482, opened 2023-08-17
Remove unsuported character by ggtext in Labels

Felixmil:
+ add utilitary function
+ add unit test

### DIFF
--- a/R/utilities-label.R
+++ b/R/utilities-label.R
@@ -94,3 +94,23 @@ getLabelWithUnit <- function(label, unit = NULL) {
     return(paste0(label, " [", unit, "]"))
   }
 }
+
+#' Sanitize Label Text
+#' @description
+#' ggtext does not allow certain characters that can be converted to html
+#  tags but that are not supported
+#'
+#'
+#' @param text a character string
+#'
+#' @return a sanitized character string
+#'
+#' @examples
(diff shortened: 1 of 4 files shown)


R/utilities-label.R
```diff
@@ -94,3 +94,23 @@ getLabelWithUnit <- function(label, unit = NULL) {
...
+#  tags but that are not supported
+#'
+#'
+#' @param text a character string
+#'
+#' @return a sanitized character string
+#'
+#' @examples
+#' sanitizeLabel("`code`")
+sanitizeLabel <- function(text){
```
PavelBal, line 110: This function should be internal, right? Then please prepend a `.` to the name.

R/utilities-label.R
```diff
@@ -94,3 +94,22 @@ getLabelWithUnit <- function(label, unit = NULL) {
     return(paste0(label, " [", unit, "]"))
   }
 }
+
+#' Sanitize Label Text
+#' @description
+#' ggtext does not allow certain characters that can be converted to html
+#  tags but that are not supported
```
PavelBal, line 101: Add that the function actually removes these characters. Also, please add a note to NEWS as this is a new behavior.

PavelBal (approved):

PR #495, opened 2023-12-12
Add exported plot test

Felixmil:
I had trouble for plot export that were fixed by updating ragg package.

In the process of looking for the source of the problem, I noticed there were no unit test for checking the result of exported plots. So I added one.


DESCRIPTION
```diff
@@ -124,3 +124,4 @@ Collate:
     'utilities-theme.R'
     'utils.R'
     'zzz.R'
+Config/testthat/edition: 3
```
Yuri05, line 127: is it the same as requiring testthat version `>=3` and `<4 `?

Felixmil: No, but equivalent to requiring version `>= 3`: https://testthat.r-lib.org/articles/third-edition.html

Yuri05: We already have `testthat (>= 3.0.3),` under `Suggests`
So what additional effect `Config/testthat/edition: 3` would have?

Felixmil: It explicitly activates the 3rd "edition" of testthat. It's a weird thing they done but for compatibility issues, but it is different than just specifying the version in `DESCRIPTION
`


If I understood correctly, it just means that you can update the package while still using previous versions of the code.

Maybe it could be an inspiration for ospsuite-R <-> PKSIM compatibility ?

Yuri05: not that I fully understand the implications, but OK :)

Yuri05 (approved):

Felixmil:
Add a skip for AppVeyor because png() device cannot be initialized correctly there.

DESCRIPTION
Felixmil: It means that there are several path for each function resolution depending on the "edition". For example, in `testthat::expect_identical`:

In edition 3 it uses `{waldo}` package while in previous editions, it uses base R.

```r
expect_identical <- function(object, expected, info = NULL, label = NULL,
                             expected.label = NULL, ...) {
   [...]

  if (edition_get() >= 3) {
    expect_waldo_equal("identical", act, exp, info, ...)
  } else {
    ident <- identical(act$val, exp$val, ...)
    if (ident) {
      msg <- ""
    } else {
      compare <- compare(act$val, exp$val)
      if (compare$equal) {
        msg <- "Objects equal but not identical"
      } else {
        msg <- compare$message
      }
    }
    [...]
}
```

And the `get_edition()` function reads from environment variables or `DESCRIPTION` file.

PR #507, opened 2024-02-27
Fixes #506

### LINKED ISSUE #506: TLF build fails
Build is currently failing because snapshot comparisons fail.
The difference between the PR whose build was successful and the new branch that failed seems to be the `{ggpot2}` version going from 3.4.4 to 3.5.0

pchelle:


pchelle:
@Yuri05 , @Felixmil 
`{ggplot2}` [version 3.5.0](https://ggplot2.tidyverse.org/news/index.html) changed things especially for the legends.
The `{ospsuite.reportingengine}` package was affected by the change (tests and build are also currently failing) and I expect that `{ospsuite}` may also be affected.

Yuri05 (approved):

PR #512, opened 2024-05-16
Fix horizontal error bars

### LINKED ISSUE #1410: After TLF update, snapshots are broken
Old:
image

New:

image

Felixmil:
Fixes https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1410

OSPSuite snapshot with this TLF version:

image


Felixmil:
The Checks keep failing because of the test of double axis time profile. Difference seems to come from graphical device or so. The image is just slighly taller...

double-axis-timeprofile

double-axis-timeprofile new

pchelle (approved):
I don't mind removing the test.
However, unfortunately, we can't deprecate/remove the feature as this type of plot is very often used in the Qualification Reports of our users (usually having concentration on the left axis and fraction on the right axis)

Felixmil:
> I don't mind removing the test. However, unfortunately, we can't deprecate/remove the feature as this type of plot is very often used in the Qualification Reports of our users (usually having concentration on the left axis and fraction on the right axis)

Do you have tests for this in the qualification framework or reporting engine ? If yes, I feel confident to remove it here.

pchelle:
> > I don't mind removing the test. However, unfortunately, we can't deprecate/remove the feature as this type of plot is very often used in the Qualification Reports of our users (usually having concentration on the left axis and fraction on the right axis)
> 
> Do you have tests for this in the qualification framework or reporting engine ? If yes, I feel confident to remove it here.

No test of the left/right axis yet. 
But good to have in [RE-Test-Reports](https://github.com/Open-Systems-Pharmacology/RE-Test-Reports) that creates and compares structures and figures of test reports on a weekly basis.

> Can you share an exemple of such a plot ?

From the [Digoxin](https://github.com/Open-Systems-Pharmacology/Digoxin-Model) qualification workflow we do have a lot of those.

18_time_profile_plot_Digoxin_Digoxin_iv__1_0_mg__1_h

PavelBal:
> I'm in favor of removing the test and deprecating the feature. Anyway, double axis plots are just bad and should be banished from this world.

Here I disagree, but I am fine with removing the test here.

PR #524, opened 2025-02-07
Increment version number

PavelBal:
Update R version
Update utils dependency
Documnetation
Styler


inst/extdata/tlf-env.RData
Yuri05, line 1: what is this?

PavelBal: Erm, no idea where this comes comes from... Delete

Yuri05 (approved):

Felixmil:
`NEWS` update is missing.
Wait until we pass checks before merging.

inst/extdata/tlf-env.RData
pchelle: If I recall, `tlfEnv` object records all the global settings of the package such as watermark, cap on error bar, etc.
The idea of `tlf-env.RData` was to enable users to [reset](https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/68df1f5bef28bf10141261cebfd011566f266d8e/R/utilities-tlf.R#L219-L226) such settings to the package default or use their own settings (`loadTLFSettings()` after saving them `saveTLFSettings()`.

pchelle (changes requested):
We can't delete the `tlf-env.RData` file

Yuri05:
> We can't delete the `tlf-env.RData` file

@PavelBal OK, let's restore the file then :)

pchelle (approved):

Yuri05 (approved):

PR #530, opened 2025-09-15
make tlf ggplot2 4.0.0 compatible

Felixmil:
# Make TLF ggplot2 4.0.0 compatible

## Summary
This PR updates the TLF package to be compatible with ggplot2 4.0.0, addressing breaking changes in the ggplot2 API while making sure it also works with ggplot 3.5.2.

## Key Changes
- **Font handling**: Updated font configuration to use proper ggplot2 4.0.0 margin syntax with `ggplot2::margin()` instead of deprecated `unit()` approach
- **Plot grid**: Fixed title and subtitle margin handling in plot grids to use the new margin specification format
- **Code formatting**: Improved code formatting and structure for better maintainability
- **Test updates**: Updated all test snapshots to reflect changes from ggplot2 4.0.0  (changes are not visible but svg are slightly different)
- **CI/CD**: Added GitHub Actions workflow for automated testing and code coverage
- **Dependencies**: Added devtools as a suggested dependency and proper remotes configuration for {osp.Rutils}
- AppVeyor is dying.


tests/testthat/_snaps/plot-obs-vs-pred/basic-with-and-without-error-bars.svg
PavelBal, line 1: Erm what happened here?

Felixmil: There is random generation of points just within the test here (L48): 

https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/ad57bbb909e002823753b634a2e8e063a6f75602/tests/testthat/test-plot-obs-vs-pred.R#L44-L66

As seed is defined at the start of the file but not directly before the test, any change in the R codebase or in other packages evaluated between this original seed and this test can have an impact on random generation.

I'll add a seed.

Yuri05:
With switching to GHA `appveyor.yml` can be safely removed

Felixmil:
@Yuri05 Done.
I also increased the patch version number.

PavelBal (approved):

PR #533, opened 2025-10-14
Fixes #532 use parse_safe to prevent crash of log ticks

### LINKED ISSUE #532: Safe parsing of log ticks
Some log tick plots crash with new ggplot2/scales versions.
See [RE Test Reports issue #110](https://github.com/Open-Systems-Pharmacology/RE-Test-Reports/issues/110) for more context

pchelle:


Yuri05 (approved):

pchelle:
> Replaced `parse()` with `scales:::parse_safe()` in three tick label generation functions

- `parse_safe` is not exported by the `{scales}` package, so I had to use `:::`
- `{ggplot2}` is a package importing `{scales}`, so it should be installed along.

PR #537, opened 2025-10-22
Fixes #536 watermark position independent from scales

### LINKED ISSUE #536: Watermark affected by scales
`ggplot2::annotation_custom()` is affected by scales in new ggplot (see [here](https://github.com/tidyverse/ggplot2/pull/6182) for more details)

This changes the behaviour of the watermarks that do not appear, especially in log scaled plots along with the following warning: `NaNs produced`

pchelle:
As suggested in this [`ggplot2` PR](https://github.com/tidyverse/ggplot2/pull/6182)

### DIFF
--- a/R/utilities-background.R
+++ b/R/utilities-background.R
@@ -270,7 +270,7 @@ addWatermark <- function(plotObject,
 
   watermark <- createWatermarkGrob(label = watermark, alpha = alpha)
 
-  plotObject <- plotObject + ggplot2::annotation_custom(grob = watermark, xmin = -Inf, ymin = -Inf, xmax = Inf, ymax = Inf)
+  plotObject <- plotObject + ggplot2::annotation_custom(grob = watermark, xmin = I(-Inf), ymin = I(-Inf), xmax = I(Inf), ymax = I(Inf))
 
   return(plotObject)
 }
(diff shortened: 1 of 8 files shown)


pchelle:
@KatrinCoboeken is this already accounted for in `{opsuite.plots}` ?

pchelle:
> @KatrinCoboeken is this already accounted for in `{opsuite.plots}` ?

I checked, you actually use `geom_text` so the fix is not relevant for the `{opsuite.plots}` package

Felixmil (changes requested):
Snapshots need to be updated too.

KatrinCoboeken:
Yes, in {ospsuite.plots} I also had to update the watermark feature with ggplot2 version 4. But it is done.

Felixmil (approved):
Plots look the same, the changes are svg "internal". All good.

Felixmil:
Should we expect plot changes in other packages ?

pchelle:
> Should we expect plot changes in other packages ?

This update should not change the plots in other packages.
Before the ggplot2 v4, `annotation_custom()` was not sensitive to scales functions, using this PR it simply remains unsensitive to scales functions

PR #544, opened 2026-07-27
Resolve OSP dependencies through the r-universe

Felixmil:
Makes the OSP R-universe the resolution channel for `tlf`, in CI and for source installs, and retires the OSP-to-OSP `Remotes:` chain for this package. `tlf` imports `ospsuite.utils`, which is not on CRAN, so resolution previously went through `Remotes:` and the GitHub API, which forced a from-source build and depended on multi-layer GitHub API resolution. It also restores this repo's CI, which had not run since April.

## CI

- `main-workflow.yaml`: pass `extra-repositories: https://open-systems-pharmacology.r-universe.dev` to the reusable `R-CMD-check-build.yaml` and `pkgdown.yaml`, which forward it to `setup-r-env` and on to `r-lib/actions/setup-r`, putting the universe on `getOption("repos")`. The `pkgdown` input comes from Workflows#249; the `R-CMD-check-build` one already existed.

## Workflow permissions

`Main-Workflow` had been ending in `startup_failure` since 2026-04-28, on `main` pushes and on PR branches, so no R check had run on this repo in three months. A called workflow cannot request a permission the caller has not granted, and the top-level grant here was only `contents: write` plus `packages: read`, while `pkgdown.yaml` requests `pull-requests: read` for its `check-changes` job (added upstream 2026-04-10) and `bump_dev_version_tag_branch.yaml` requests `actions: read`. Granting those at job level rather than widening the top-level block restores the run. Pre-existing and unrelated to the universe work, fixed here because otherwise nothing else in this PR can be validated on this repo.

## DESCRIPTION

- Declare `Additional_repositories: https://open-systems-pharmacology.r-universe.dev`, so `R CMD check` can validate the non-CRAN dependency and source installs (`install.packages(dependencies = TRUE)`, `remotes::install_local()` on a tarball) can find `ospsuite.utils` without pak.
- Drop the `Remotes:` entry for `ospsuite.utils`. This is the change that actually switches the channel, and the additive steps alone do nothing without it. Measured on this branch: with `Remotes:` present and the universe already on `repos`, pak ignored the universe and fetched `ospsuite.utils 1.11.1.9001` as a GitHub zipball. With the field removed it resolves `ospsuite.utils 1.11.1` from `https://open-systems-pharmacology.r-universe.dev/src/contrib/`. `Imports` already required `>= 1.6.0`, which the universe release satisfies, so no version floor change was needed.

The dependency change is proven independently of the test failures below: `pkgdown` is green, and `R-CMD-check` passes on Windows, both resolving `ospsuite.utils` from the universe with no `Remotes:` field present.

## Outstanding: vdiffr snapshots are platform-dependent

`R-CMD-check` is red on Linux and macOS from visual snapshot comparisons, not from anything in the dependency change. After the snapshot refresh in this branch the results are: Windows passes, ubuntu-22.04 reports `FAIL 1` (`test-pkratio.R:79:3`), macos-15 reports `FAIL 4` (`test-font-label.R:178:3`, `test-pkratio.R:79:3`, `test-plot-grid.R:47:3`, `test-plot-grid.R:104:3`). vdiffr renders SVG through the platform font stack, so one accepted set of snapshots cannot satisfy three different runners, and refreshing them on one platform is what produced this split. Chasing per-platform snapshots is a treadmill; the durable fix is to run the visual tests on a single reference platform, via `skip_on_os()` around the `expect_doppelganger()` calls or a matrix that runs vdiffr on one runner only. That is test-infrastructure work worth separating from this PR.

The snapshot commit also carries files that look accidental and should come off before merge: several `*.new.svg` and one `*.new.png` (vdiffr writes these when a comparison fails, they are not accepted snapshots), a modified `inst/extdata/tlf-env.RData`, and a stray `tests/testthat/test-export-obs-vs-pred.PNG` at the top of `tests/testthat/`.

## README

Rewrite the installation section around the universe. The released version now installs with plain base R and no extra tooling:

```r
install.packages(
  "tlf",
  repos = c(OSP = "https://open-systems-pharmacology.r-universe.dev", getOption("repos"))
)
```

The development version keeps a documented path via `pak::pak("Open-Systems-Pharmacology/TLF-Library")`. The instructions for installing a downloaded binary are removed: `install.packages()` ignores `dependencies` when `repos = NULL`, so that path can never resolve dependencies on its own, and the old text handled this with a hand-maintained list of `install.packages()` calls plus a link to `ospsuite.utils_1.3.17.zip`, a version that no longer satisfies this package's own `Imports` floor. Since the universe serves binaries for Windows and macOS, the command above covers what that section was for.

This covers the CI and `DESCRIPTION` steps plus the `Remotes:` retirement from the plan tracked in the R-universe registry issue "Wire OSP dependency resolution through R-universe": https://github.com/Open-Systems-Pharmacology/open-systems-pharmacology.r-universe.dev/issues/2

## Summary by CodeRabbit

* **Documentation**
  * Updated installation instructions for the released package using the OSP R-universe.
  * Added streamlined instructions for installing the development version from GitHub.
  * Removed outdated manual installation guidance.

* **Bug Fixes**
  * Improved consistency of rendered plots and snapshots across Windows, macOS, and Linux.
  * Stabilized plot legend ordering and platform-specific visual comparisons.


pchelle (approved):

inst/extdata/tlf-env.RData
Yuri05, line 1: what was changed in **tlf-env.RData**?

Felixmil: Nothing meaningful, and it is now reverted (2f2ed87), so this file no longer appears in the PR diff.

No setting value had changed. The only difference was absolute filesystem paths from my local checkout, which end up in the file as `srcref` attributes on the functions stored in `currentTheme` and `defaultAggregation`. That accounted for the whole 66-byte delta.

Root cause: the `save()` at the bottom of `R/utils.R` is top-level code, so it also runs on `devtools::load_all()`, which regenerates the tracked `inst/extdata/tlf-env.RData` on every dev session. Dev sessions parse with `keep.source = TRUE` (installs do not), which is where the `srcref`s and their embedded paths come from. Very easy to commit by accident. Restricting that `save()` to real installs fixes it.

Yuri05: Ok, thx for the clarification

Yuri05 (approved):

Yuri05 (approved):

